PR #5, opened 2019-02-28
AISDK-9: Create package for node sdk

kbridbur:


src/api-client.ts
```diff
@@ -13,10 +13,7 @@ import RevAiApiJob from './models/RevAiApiJob';
 import RevAiApiTranscript from './models/RevAiApiTranscript';
 import RevAiJobOptions from './models/RevAiJobOptions';
 
-/* tslint:disable-next-line:no-var-requires */
-const versionNumber = require('../package.json').version;
```
kbridbur: This import isn't possible with the current structure of the package. I can explain more in detail if anyone wants but currently it would require rearranging the package in a way which I don't think is intuituve

.travis.yml
```diff
@@ -1,6 +1,8 @@
 {
   "language": "node_js",
-  "node_js": "8",
+  "node_js": 
+    - "8"
+    - "10",
```
amikofalvy: We can experiment with other versions of nodejs in the future, but these are good for now.

JenkinsFile
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+    environment {
+        GIT_CREDENTIALS = credentials("${GITHUB_CREDENTIALS}")
+    }
+    stages {
+        stage('Lint/Test') {
+            steps {
+                sh """
+                    npm i
```
amikofalvy: Please use the full `npm install` command

JenkinsFile
```diff
@@ -0,0 +1,129 @@
...
+                    def data = readJSON file:'package.json'
+                    checkVersionAlreadyExists(data.version)
+                }
+            }
+        }
+
+        stage('Publish') {
+            steps {
+                sh """
+                    npm run build
```
amikofalvy: Should we separate out the build step? Build is a distinct item separate from publish.

kbridbur: I think in this case they're not as separate. "Build" in this case is really just "Transpile to Js"

JenkinsFile
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+        }
+    }
+    post {
+        success {
+            notifyComplete('SUCCESS')
+        }
+        failure {
+            notifyComplete('FAILURE')
```
amikofalvy: Let's copy the logic from RevAiApi and RevCommon for not showing failure messages if the previous build was a failure. This will reduce bot spam in our development slack channel.

JenkinsFile
kbridbur: I guess I should change the script name to match this. It should be `npm run transpile`

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@rev_ai/revai-node-sdk",
-  "version": "1.0.0",
+  "name": "revai-node-sdk",
+  "version": "1.0.2",
```
amikofalvy: I think we're on version 1.0.4 now, we'll need to increase the version, if we want to deploy again.

package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "revai-node-sdk",
-  "version": "1.0.0",
+  "version": "1.0.5",
   "description": "Rev.ai makes speech applications easy to build!",
   "main": "index.js",
   "scripts": {
-    "build": "tsc",
+    "transpile": "tsc",
```
amikofalvy: build is the convention

package-lock.json
amikofalvy: Actually, I didn't realize this was the package-lock. Why is this included the package lock file?

kbridbur: Package-lock doesn't actually get transferred over when we deploy. I'm not sure if we can fully remove the file but its effectively dead code. Only the version in package.json matters

amikofalvy: Usually this is done by npm cli when you run npm uninstall package-name. I think you should try running this command to fix the package-lock

PR #6, opened 2019-03-06
 AISDK-33: Catch Node SDK up to current API functionality

k-weng:


README.md
```diff
@@ -55,9 +55,38 @@ var jobDetails = client.getJobDetails(job.id);
 `jobDetails` will contain all information normally found in a successful response from
 our [Get Job](https://www.rev.ai/docs#operation/GetJobById) endpoint
 
+### Checking multiple files
+
+You can retrieve a list of transcription jobs with optional parameters
+
+```javascript
+var jobs = client.getListOfJobs();
```
dkokotov: I had mentioned this to @kbridbur - I think these (and other examples in the README) should be `await client.[whatever])`. since the methods return promises

kbridbur: This is my fault but I committed a change directly to master earlier today and forgot to push the rebase of develop. All the methods in the docs should be awaited. So this should be `var jobs = await client.getListOfJobs();`

test/api-client.spec.ts
```diff
@@ -31,155 +33,267 @@ describe('api-client', () => {
...
             }
         });
     });
-    
+
+    describe('getListOfJobs', () => {
+        afterEach(() => {
+            expect(mockedAxios.get).toBeCalledTimes(1);
+        })
+
```
kbridbur: I would add a test with no parameters

PR #8, opened 2019-05-16
AISDK-36: Add streaming client to sdk

kbridbur:
Adds streaming functionality to the node sdk

Includes an example in streaming-client.ts

TODO:
- [x] : Add connection tests


src/streaming-client.ts
```diff
@@ -0,0 +1,90 @@
...
+/*
+Example: 
+let audioConfig = new AudioConfig();
+audioConfig.contentType = "audio/x-wav";
+const token = "token";
+var client = new RevAiStreamingClient(token, audioConfig);
+var stream = client.start();
+var file = fs.createReadStream('../test.wav');
+file.pipe(stream);
+stream.on('data', data => {
```
kbridbur: `on('readable', () => {
    console.log(stream.read());
});`
Also works

src/streaming-client.ts
```diff
@@ -0,0 +1,90 @@
...
+        });
+        this.client.on('connect', function(connection: any) {
+            connection.on('error', function(error : Error) {
+                console.log("Error thrown in connection to streaming service");
+                self.requests.end();
+                self.responses.push(null);
+                throw error;
+            });
+            connection.on('close', function(code: number, reason: string) {
+                console.log(`Connection closed, ${code}: ${reason}`);
```
kbridbur: I'm debating what the best way to return this information to the user is. I have 3 potential solutions
1. RevAiStreamingClient inherits from WebSocketClient so all these events can be thrown and seen by the user.
2. Store close status as a property
3. Write a close message to the output stream.

I personally think #1 is the best since it gives the user the most flexibility but I wanted to hear other opinions

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+    channels?: number;
+    constructor(contentType: string, layout?: string, rate?: number, format?: string, channels?: number){
+        this.contentType = contentType;
+        this.layout = layout;
+        this.rate = rate;
+        this.format = format;
+        this.channels = channels;
+    }
+
+    getContentTypeString (): string {
```
kennygea: you have a whitespace between the `function` and `()`, I don't think this is correct TS standard? 

btw I'm following https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#variable-and-function

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+        return `${this.contentType}` +
+            (this.layout == null ? "" : `;layout=${this.layout}`) +
```
kennygea: do you have to check for `null`? Can `this.layout` be `undefined` as well? Would a check like `this.layout ? "" : `;layout=${this.layout}`` work as well?

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,22 @@
+export default class AudioConfig {
```
kennygea: Some js docs would be nice

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+            return this.setUpOutputHandlersAndRead(size);
+        }
+        return this.readOutput(size);
+    }
+
+    setUpOutputHandlersAndRead (size : any): void {
+        var self = this;
+        self.outputHandlersSetup = true;
+        self.output
+        .on('readable', function () {
```
kennygea: this formatting seems off, `.on` seems like it should be `indented`

src/streaming-client.ts
```diff
@@ -0,0 +1,97 @@
...
+import StreamingResponse from './models/streaming/StreamingResponse';
+
+export class RevAiStreamingClient extends events.EventEmitter {
+    accessToken: string;
+    config: AudioConfig;
+    baseUrl: string;
+    requests: PassThrough;
+    responses: PassThrough;
+    client: sockets.WebSocketClient;
+    constructor(accessToken: string, config: AudioConfig, version = 'v1alpha'){
```
kennygea: couldn't `v1alpha` just be a private const on the class?

src/streaming-client.ts
```diff
@@ -0,0 +1,97 @@
...
+        super();
+        this.accessToken = accessToken;
+        this.config = config;
+        this.baseUrl = `wss://api.rev.ai/speechtotext/${version}/stream?` +
+            `access_token=${accessToken}` +
+            `&content_type=${config.getContentTypeString()}`;
```
kennygea: I think we should either do some validation `AudioConfig` client side, or at least provide defaults. What do you think?

src/streaming-client.ts
```diff
@@ -0,0 +1,97 @@
...
+        this.client.on('connectFailed', function(error: Error) {
+            self.emit('connectFailed', error);
+            self.requests.end();
+        });
+        this.client.on('connect', function(connection: any) {
+            self.emit('connect');
+            connection.on('error', function(error : Error) {
+                self.emit('error', error);
+                self.requests.end();
+                self.responses.push(null);
```
kennygea: just curious, what is this `push(null)` doing?

src/streaming-client.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    start (): Duplex {
+        var self = this;
+        this.client.on('connectFailed', function(error: Error) {
+            self.emit('connectFailed', error);
+            self.requests.end();
+        });
+        this.client.on('connect', function(connection: any) {
+            self.emit('connect');
```
kennygea: shouldn't we return the `ConnectedStreamingResponse` here? It contains `JobId` which I imagine is crucial for users

src/streaming-client.ts
kennygea: I guess this is how we do it for the regular client.

kennygea: missing space between `)` and `{` at the end here

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    _read (size: any): any {
+        if (!this.outputHandlersSetup){
+            return this.setUpOutputHandlersAndRead(size);
+        }
+        return this.readOutput(size);
+    }
+
+    setUpOutputHandlersAndRead (size : any): void {
```
kennygea: private?

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+        var self = this;
+        self.outputHandlersSetup = true;
```
kennygea: minor, to follow the flow of the `outputHandlersSetup` I would move this assignment to after the handlers have been set

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export default class BufferedDuplex extends Duplex {
+    input: PassThrough;
+    output: PassThrough;
+    outputHandlersSetup: boolean;
+    constructor (input: PassThrough, output: PassThrough, options?: any) {
+        super(options);
+        this.input = input;
+        this.output = output;
+        this.outputHandlersSetup = false;
```
kennygea: boolean name, maybe should be `isOutputHandlersSetup`, otherwise the naming feels like a function

src/streaming-client.ts
kennygea: also these messages should be the past tense, so like `self.emit('connected')` and `self.emit('closed')` down below etc

src/streaming-client.ts
```diff
@@ -0,0 +1,97 @@
...
+                self.requests.end();
+                self.responses.push(null);
+            });
+            connection.on('message', function(message: any) {
+                if (message.type === 'utf8') {
+                    self.responses.write(JSON.parse(message.utf8Data) as StreamingResponse);
+                }
+            });
+            
+            function sendFromBuffer() {
```
kennygea: personal preference, but you could make this not a local function, by having `sendFromBuffer(connection: any)` and pass the connection through

src/streaming-client.ts
kennygea: also `access_token`, right now the backend returns a 401 (or 403?) http status code on unauthorized. Can we support returning the proper error in this client?

src/models/streaming/AudioConfig.ts
kbridbur: It can be null, == covers both of these cases whereas === would only cover if it was exactly null. I'll see if your suggestion works, that would be cleaner

src/streaming-client.ts
kbridbur: I think defaulting to `audio/*` is reasonable. I don't think doing validation is reasonable for the same reason we don't validate on the midtier. It would be nice in order to fail fast but it isn't the responsibility of this sdk imo

src/models/streaming/AudioConfig.ts
kennygea: I know we can do `!this.layout ?` for sure

src/streaming-client.ts
kbridbur: It ends the output stream so the 'end' event is triggered

src/streaming-client.ts
kbridbur: That response is also read from the responses output stream currently. I agree that it would be a better user experience if this returned it instead.

The standard for event names in javascript seems to be present tense. I can't find it written anywhere but based off the events emitted by streams and websockets

src/models/streaming/AudioConfig.ts
kbridbur: It works. I wasn't sure if `this.layout` on its own would handle undefined checking but it makes sense that it does

src/models/streaming/AudioConfig.ts
kennygea: bump on jsdocs

src/streaming-client.ts
```diff
@@ -0,0 +1,115 @@
...
+                self.responses.push(null);
+            });
+            connection.on('close', function(code: number, reason: string) {
+                self.emit('close', code, reason);
+                self.end();
+            });
+            connection.on('message', function(message: any) {
+                if (message.type === 'utf8') {
+                    var response = JSON.parse(message.utf8Data) as StreamingResponse;
+                    if (response.type == "connected"){
```
kennygea: space between `)` and `{`

src/streaming-client.ts
```diff
@@ -0,0 +1,115 @@
...
+                        self.emit('connect', response.id);
+                    }
+                    else{
```
kennygea: space after `else`

src/streaming-client.ts
```diff
@@ -0,0 +1,115 @@
...
+            });
+            connection.on('close', function(code: number, reason: string) {
+                self.emit('close', code, reason);
+                self.end();
+            });
+            connection.on('message', function(message: any) {
+                if (message.type === 'utf8') {
+                    var response = JSON.parse(message.utf8Data) as StreamingResponse;
+                    if (response.type == "connected"){
+                        self.emit('connect', response.id);
```
kennygea: I would actually return the whole `StreamingResponse` and maybe split up `StreamingResponse` into 2 classes (or 3, with 1 as a base). My reasoning is it's unclear what to parse for using the SDK as it currently is for each response. 

If we include `type` back in every response, it's easier to document and code-wise to check which message is what

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
+import { Duplex, PassThrough } from 'stream';
+
+export default class BufferedDuplex extends Duplex {
+    input: PassThrough;
+    output: PassThrough;
+    areOutputHandlersSetup: boolean;
+    constructor (input: PassThrough, output: PassThrough, options?: any) {
```
kennygea: does this `options` value have a type?

src/streaming-client.ts
```diff
@@ -0,0 +1,115 @@
...
+                };
+            });
+            
+            function sendFromBuffer() {
+                if (connection.connected) {
+                    var value = self.requests.read(self.requests.readableLength);
+                    if (value != null) {
+                        connection.send(value);
+                    }
+                    setTimeout(sendFromBuffer, 100);
```
kennygea: is there a specific reasoning for a `100` millisecond timeout? If so, I would document it

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+            })
+            .on('end', function () {
+                self.push(null);
+            });
+        self.areOutputHandlersSetup = true;
+    }
+
+    private readOutput(size : any): void {
+        var chunk;
+        while(null !== (chunk = this.output.read(size))) {
```
kennygea: super minor, should this boolean check be flipped? Seems kind of weird to have `null` first

src/streaming-client.ts
kbridbur: No specific reason for 100. This value is still up for debate. I'll likely do some latency testing with different values to find an ideal one since I'm not sure what the internals of `pipe` look like

src/models/streaming/BufferedDuplex.ts
kbridbur: https://nodejs.org/api/stream.html#stream_new_stream_transform_options It does not according to this documentation

kbridbur:
@kennygea re: docs 
I have a separate issue specced to add docs to the readme. I would prefer to wait until after we add streaming to our normal documentation since the SDK docs often link to parts of the API docs

src/streaming-client.ts
```diff
@@ -0,0 +1,117 @@
+import * as fs from 'fs';
+import * as sockets from 'websocket';
+import * as events from 'events';
+import { Duplex, PassThrough } from 'stream';
+
+import AudioConfig from './models/streaming/AudioConfig';
```
kennygea: minor, I would always use `{ } ` after `import` even it it's singular.

src/streaming-client.ts
kennygea: thoughts on this comment?

src/streaming-client.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    end(): void {
+        this.client.abort();
+        this.requests.end("EOS");
+        this.responses.push(null);
+    }
+
+    private setUpClient(): void {
+        var self = this;
+        this.client.on('httpResponse', function(response, client) {
```
kennygea: minor, I would move each of these `this.client.on` into separate private helper functions. 

so `setupHttpResponseHandler`, `setupConnectionFailedHandler`, `setupConnectHandler`, etc

src/streaming-client.ts
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+            sendFromBuffer();
+        });
+    }
+}
+
+/*
+// Example: 
+let audioConfig = new AudioConfig("audio/x-wav");
```
kennygea: I don't know if this should necessarily go here. Maybe a separate `example` file, or even a separate `.md` file

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,188 @@
+import { PassThrough } from 'stream';
+import { WebSocketClient, WebSocketConnection } from 'websocket';
+import AudioConfig from '../src/models/streaming/AudioConfig';
```
kennygea: ditto on the `import { }` here

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+            })
+            .on('end', function () {
+                self.push(null);
+            });
+        self.areOutputHandlersSetup = true;
+    }
+
+    private readOutput(size : any): void {
+        var chunk;
+        while((chunk = this.output.read(size)) !== null) {
```
kennygea: more minor, I would just do ` var chunk = this.output.read(size)` in the first line. it's basically the same thing

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+    contentType: string;
+    layout?: string;
+    rate?: number;
+    format?: string;
+    channels?: number;
+    constructor(contentType = "audio/*", layout?: string, rate?: number, format?: string, channels?: number){
```
kennygea: space here `) {`

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+        if (!this.areOutputHandlersSetup){
+            return this.setUpOutputHandlersAndRead(size);
+        }
+        return this.readOutput(size);
+    }
+
+    private setUpOutputHandlersAndRead(size : any): void {
+        var self = this;
+        self.output
+            .on('readable', function () {
```
kennygea: I believe using `() =>` arrow functions preserves `this` through the calls, so you don't have to do `var self = this` for when you want to pass `this` through to callbacks. 

https://hackernoon.com/javascript-es6-arrow-functions-and-lexical-this-f2a3e2a5e8c4

I also think this is standard typescript practice for us, so I would update to use `() =>` as much as possible

src/streaming-client.ts
kbridbur: I prefer to keep it within the connection. I view it as a function which is part of the connection not a part of the client as a whole.

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,22 @@
+export class AudioConfig {
```
amikofalvy: please add js docs

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,22 @@
...
+    contentType: string;
+    layout?: string;
+    rate?: number;
+    format?: string;
+    channels?: number;
+    constructor(contentType = "audio/*", layout?: string, rate?: number, format?: string, channels?: number) {
```
amikofalvy: minor space
```suggestion
    
    constructor(contentType = "audio/*", layout?: string, rate?: number, format?: string, channels?: number) {
```

amikofalvy: What do you think about this method of private property definition?
 https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/

The syntax is not popular, so it might be too confusing.

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
+import { Duplex, PassThrough } from 'stream';
+
+export class BufferedDuplex extends Duplex {
```
amikofalvy: please put docs on every class.

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+    input: PassThrough;
+    output: PassThrough;
```
amikofalvy: docs on properties are helpful too, but only when required.

src/models/streaming/BufferedDuplex.ts
amikofalvy: Kenny, is correct. Using arrow functions everywhere can help us reduce the number of scoping errors we accidentally introduce.

The mobile team has implemented an arrow function only policy and has seen a decrease in scoping issues.

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,43 @@
...
+        this.output = output;
+        this.areOutputHandlersSetup = false;
+    }
+
+    _write(chunk: any, encoding: any, callback: any): boolean {
+        return this.input.write(chunk, encoding, callback);
+    }
+
+    _read(size: any): any {
+        if (!this.areOutputHandlersSetup){
```
amikofalvy: I'm not familiar with how exactly `Duplex` operates. If we setup outputhandlers only when _read is invoked are there cases when this would be too late to read incoming messages?



also minor spacing edit
```suggestion
        if (!this.areOutputHandlersSetup) {
```

src/streaming-client.ts
```diff
@@ -0,0 +1,156 @@
...
+    StreamingHypothesis, 
+    StreamingConnected 
+} from './models/streaming/StreamingResponses';
+
+/**
+ * Client which handles a streaming connection to the Rev.ai API.
+ *
+ * @param accessToken {string} Access token associated with the user's account
+ * @param config {AudioConfig} Configuration of the audio the user will send from this client
+ * @param version (optional) {string} Version of the Rev.ai API the user wants to use
```
amikofalvy: What is the industry convention for where to put docs for constructor parameters? Do we put them by the constructor definition?

src/streaming-client.ts
```diff
@@ -0,0 +1,156 @@
...
+ * file.pipe(stream);
+ */
+export class RevAiStreamingClient extends events.EventEmitter {
+    baseUrl: string;
+    client: sockets.WebSocketClient;
+    private accessToken: string;
+    private config: AudioConfig;
+    private requests: PassThrough;
+    private responses: PassThrough;
+    constructor(accessToken: string, config: AudioConfig, version = 'v1alpha'){
```
amikofalvy:
```suggestion
    
    constructor(accessToken: string, config: AudioConfig, version = 'v1alpha'){
```

amikofalvy: Please run a linter on these files.

src/streaming-client.ts
amikofalvy: 👍 Don't do validation on the client side as this duplicates validation logic and slows down our development.

src/streaming-client.ts
```diff
@@ -0,0 +1,156 @@
...
+        this.accessToken = accessToken;
+        this.config = config;
+        this.baseUrl = `wss://api.rev.ai/speechtotext/${version}/stream?` +
+            `access_token=${accessToken}` +
+            `&content_type=${config.getContentTypeString()}`;
+        this.requests = new PassThrough();
+        this.responses = new PassThrough({ objectMode: true });
+        this.client = new sockets.client();
+    }
+    /**
```
amikofalvy:
```suggestion
    
    /**
```

src/streaming-client.ts
```diff
@@ -0,0 +1,156 @@
...
+ * 
+ * var file = fs.createReadStream(<YOUR-AUDIO-FILE>);
+ * stream.on('data', data => {
+ *     console.log(data);
+ * });
+ * stream.on('end', function () {
+ *     console.log("End of Stream");
+ * });
+ * 
+ * file.pipe(stream);
```
amikofalvy: This example might have better visibility in a `Getting Started` section of the readme.

test/mocks/websocket-mock.ts
```diff
@@ -0,0 +1,25 @@
+import * as events from 'events';
+
+export class WebSocketConnection extends events.EventEmitter {
```
amikofalvy:
```suggestion
export class WebSocketConnectionMock extends events.EventEmitter {
```

amikofalvy: same with the other classes in this mock file

amikofalvy: You may need to map the names from `WebSocketConnectionMock` to `WebSocketConnection` in order for the mocks to replace the originals.

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+        sut.client = mockClient;
+    });
+
+    describe('start', () => {
+        it('Connects to api with parameters', () => {
+            const res = sut.start();
+
+            expect(mockClient.connect).toBeCalledWith(`wss://api.rev.ai/speechtotext/v1alpha/stream?` +
+                `access_token=${token}` +
+                `&content_type=${audioConfig.getContentTypeString()}`
```
amikofalvy: Another way to write this test that decouples dependency on any implementation of getContentTypeString:
```
// Setup
const fakeAudioContentString = 'fakeAudioContentString';
audioConfig.getContentTypeString = jest.fn();
audioConfig.getContentTypeString.mockReturnValue(fakeAudioContentString);
...
// Assert
expect(audioConfig.getContentTypeString).toHaveBeenCalled();
...
`&content_type=${fakeAudioContentString}`
...
```

amikofalvy: I think we should write our ts tests like this from now on.

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+            });
+            
+            mockClient.emit('connectFailed', expectedError);
+
+            expect(connectionError).toBe(expectedError);
+        });
+
+        it('Emits close code and reason on connection close', () => {
+            const res = sut.start();
+            var closeCode = null;
```
amikofalvy: use `let` instead of `var`

amikofalvy: `var` is function scoped `let` is block scoped

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+        });
+
+        it('Returns duplex stream', () => {
+            const res = sut.start();
+
+            expect(res).toBeInstanceOf(BufferedDuplex);
+        });
+    });
+
+    describe('client setup', () => {
```
amikofalvy: I think client setup tests should be a part of `start` tests.

We shouldn't have tests for private functions.

amikofalvy: You can rephrase the test names "addsEventListener to httpResponse", etc.

src/models/streaming/AudioConfig.ts
kbridbur: I'm not a huge fan. It would remove some lines but I think it would make it a lot less clear to anyone who doesn't know this syntax

src/models/streaming/BufferedDuplex.ts
kbridbur: I think what you're saying is possible. But I would blame the user in that case. If they haven't set up to read from the output buffer before it gets totally full then it's on them. I'm also not sure what a good solution to this would be. Since setting it up requires reading from it in the first place

src/streaming-client.ts
kbridbur: It actually looks like google documents these on the constructor method itself. I'll copy that convention

src/streaming-client.ts
kbridbur: I don't plan on writing the readme until we add streaming documentation to the website since it will definitely need to reference that. Since this isn't meant to be very visible until then I think keeeping the example here is good

src/api-client.ts
```diff
@@ -2,7 +2,8 @@ import axios, { AxiosInstance } from 'axios';
 import * as FormData from 'form-data';
 import * as fs from 'fs';
 
-import RevAiAccount from './models/RevAiAccount';
+import RevAiAccount from './models/async/RevAiAccount';
+import RevAiJobOptions from './models/async/RevAiJobOptions';
```
amikofalvy: can we use the explicit import instead of the default import
```suggestion
import { RevAiJobOptions } from './models/async/RevAiJobOptions';
```

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,39 @@
...
+    rate?: number;
+    format?: string;
+    channels?: number;
+
+    /**
+     * @param contentType {string} Content type of the audio stream.
+     * @param layout (optional) {string} Layout of the audio channels. "interleaved" or "non-interleaved".
+     * @param rate (optional) {int} Sample rate of audio. 8000-48000hz supported.
+     * @param format (optional) {string} Format of audio.
+     * @param channels (optional) {int} Number of audio channels. 1-10 supported.
```
amikofalvy: Does google document type?

I find it strange that we document types for two reasons:
1. int type does not explicitly exist in JS, but if Google is doing this in their docs, I'm fine with doing it in ours.
2. We already have type definitions as part of typescript. Do documentation comments end up in the transpiled js files found in the distribution?

tslint.json
```diff
@@ -112,7 +112,6 @@
               "singleline": "never"
           }
       ],
-      "space-before-function-paren": true,
```
amikofalvy: can you make ts lint a blocking step in the build so that travis ci will fail when ts lint fails?

I think you can add it to package.json

```
"scripts": {
    "build": "tslint && ..."
}
```

src/models/streaming/BufferedDuplex.ts
kbridbur: re: this conversation I realized they're not the same thing. The way it was written chunk was reevaluated each round of the while loop. If I move it to the first line then it is not

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
...
+                    let response = JSON.parse(message.utf8Data);
+                    if ((response as StreamingResponse).type === 'connected') {
+                        this.emit('connect', response as StreamingConnected);
+                    } else {
+                        this.responses.write(response as StreamingHypothesis);
+                    }
+                }
+            });
+
+            function sendFromBuffer(buffer: PassThrough): void {
```
amikofalvy: you can use arrow functions to define all instance functions

moving the function definition out of the on 'connect' callback will make this a bit cleaner.

sendFromBuffer = (buffer: PassThrough): Void => {
}

src/models/streaming/AudioConfig.ts
kbridbur: 1. You're right, I should have been using number. 
2. That's fair, we do include the .d.ts files in the distribution. Google does not write in typescript so that's most likely why they include types. I will remove them

tslint.json
kbridbur: I think it makes more sense to make it a step in the jenkins job. I'll add it there

amikofalvy: If it's only a step in travis-ci then a developer may not see a tslint error until it hits ci

package.json
```diff
@@ -33,8 +33,9 @@
   "homepage": "https://github.com/revdotcom/revai-node-sdk#readme",
   "dependencies": {
     "axios": "^0.18.0",
-    "express": "^4.16.4",
-    "form-data": "^2.3.3"
+    "form-data": "^2.3.3",
+    "stream": "0.0.2",
```
amikofalvy: if we are using nodejs streams then we shouldn't include this dependency to a third party stream library.

package.json
```diff
@@ -33,8 +33,8 @@
   "homepage": "https://github.com/revdotcom/revai-node-sdk#readme",
   "dependencies": {
     "axios": "^0.18.0",
-    "express": "^4.16.4",
-    "form-data": "^2.3.3"
+    "form-data": "^2.3.3",
```
amikofalvy: make sure to run npm install locally and add the changes to package-lock.json to match any updates made to dependencies

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,39 @@
+/**
```
amikofalvy: This comment pertains to RevAiAccount file above. Github doesn't let me leave a comment on rename only files.

Are we certain that RevAiAccount.ts should be placed into async subdirectory? It's not part of async speech.

src/models/streaming/AudioConfig.ts
```diff
@@ -0,0 +1,39 @@
...
+        channels?: number
+    ) {
+        this.contentType = contentType;
+        this.layout = layout;
+        this.rate = rate;
+        this.format = format;
+        this.channels = channels;
+    }
+
+    getContentTypeString(): string {
```
amikofalvy: please use `public` and `private` access modifiers to clarify isntance functions

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,53 @@
...
+     * @param options Options to be passed through to the superclass.
+     */
+    constructor (input: PassThrough, output: PassThrough, options?: any) {
+        super(options);
+        this.input = input;
+        this.output = output;
+        this.areOutputHandlersSetup = false;
+    }
+
+    _write(chunk: any, encoding: any, callback: any): boolean {
```
amikofalvy: should we put types on some of these parameters?
* chunk?
* encoding 'string'
* callback as `any` is perfectly fine

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,53 @@
...
+        return this.input.write(chunk, encoding, callback);
+    }
+
+    _read(size: any): any {
```
amikofalvy: size: number?

src/models/streaming/BufferedDuplex.ts
```diff
@@ -0,0 +1,53 @@
...
+        if (!this.areOutputHandlersSetup) {
+            return this.setUpOutputHandlersAndRead(size);
+        }
+        return this.readOutput(size);
+    }
+
+    private setUpOutputHandlersAndRead(size: any): void {
```
amikofalvy: propagte type

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
...
+    /**
+     * @param accessToken Access token associated with the user's account
+     * @param config Configuration of the audio the user will send from this client
+     * @param version (optional) Version of the Rev.ai API the user wants to use
+     */
+    constructor(accessToken: string, config: AudioConfig, version = 'v1alpha') {
+        super();
+        this.accessToken = accessToken;
+        this.config = config;
+        this.baseUrl = `wss://api.rev.ai/speechtotext/${version}/stream?` +
```
amikofalvy: This same question applies to both async and streaming. I think we should make baseUrl configurable so it can be pointed at other environments. Default environment would be the production environment.
Was there prior discussion on this?

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Sets up the client and begins the streaming connection. Returns a duplex
+     * from which the user can read responses from the api and to which the user
+     * should write their audio data
+     *
+     * @returns BufferedDuplex. Data written to this buffer will be sent to the api
+     * Data received from the api can be read from this duplex
+     */
+    start(): Duplex {
```
amikofalvy: public access modifer for clarification

test/setup.ts
```diff
@@ -0,0 +1,7 @@
+jest.mock('websocket', () => {
+    return {
+        WebSocketClient: require('./mocks/websocket-mock').WebSocketClientMock,
+        WebSocketConnection: require('./mocks/websocket-mock').WebSocketConnectionMock,
```
amikofalvy: 👍

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
+import * as events from 'events';
+import * as fs from 'fs';
+import { Duplex, PassThrough } from 'stream';
+import * as sockets from 'websocket';
```
amikofalvy: you can use explicit import of { client, WebSocketClient, ... }

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
+import * as events from 'events';
```
amikofalvy:
```suggestion
import { EventEmitter } from 'events';
```

test/mocks/websocket-mock.ts
```diff
@@ -0,0 +1,21 @@
+import * as events from 'events';
```
amikofalvy:
```suggestion
import { EventEmitter } from 'events';
```

test/streaming-client.spec.ts
amikofalvy: Did you decide on wether to implement my suggested refactor?

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,178 @@
+import { PassThrough } from 'stream';
```
amikofalvy: Is this import used?

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+import { WebSocketClient, WebSocketConnection } from 'websocket';
+import { AudioConfig } from '../src/models/streaming/AudioConfig';
+import { BufferedDuplex } from '../src/models/streaming/BufferedDuplex';
+import { RevAiStreamingClient } from '../src/streaming-client';
+const fs = require('fs');
+const events = require('events');
+
+jest.useFakeTimers();
```
amikofalvy: are we using this?

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
...
+    end(): void {
+        this.client.abort();
+        this.requests.end('EOS');
+        this.responses.push(null);
+    }
+
+    private setUpClient(): void {
+        this.setUpHttpResponseHandler();
+        this.setUpConnectionFailureHandler();
+        this.setUpConnectedHandlers();
```
amikofalvy: I think we can move these 3 lines directly into the beginning of start()

test/streaming-client.spec.ts
kbridbur: Yes, we call runOnlyPendingTimers on line 150

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private setUpConnectionFailureHandler(): void {
+        this.client.on('connectFailed', (error: Error) => {
+            this.emit('connectFailed', error);
+            this.end();
+        });
+    }
+
+    private setUpConnectedHandlers(): void {
```
amikofalvy:
```suggestion
    private setUpConnectionHandlers(): void {
```

src/models/streaming/AudioConfig.ts
kbridbur: I added it here since it's only relevant to async responses. I can see an argument for either location

src/streaming-client.ts
kbridbur: Yes there was. We don't expose it as publicly configurable since that makes no sense to customers. But I leave it as not private so that we can assign it if we want to. 

ie we can do
`let client = new RevAiStreamingClient()`
`client.baseUrl = 'some other environment'`

kbridbur: I guess it's a bit harder with this one as the access token and such are baked into the url. But I still think this is the best solution

src/models/streaming/BufferedDuplex.ts
kbridbur: Chunk has no type since it can be either a buffer or not

package.json
kbridbur: I've been using npm uninstall to get rid of packages so package-lock is up to date. :)

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
...
+            connection.on('error', (error: Error) => {
+                this.emit('error', error);
+                this.end();
+            });
+            connection.on('close', (code: number, reason: string) => {
+                this.emit('close', code, reason);
+                this.end();
+            });
+            connection.on('message', (message: any) => {
+                if (message.type === 'utf8') {
```
amikofalvy: what should happen if the message type is not utf8?

amikofalvy: Why do we make this call?

src/streaming-client.ts
```diff
@@ -0,0 +1,157 @@
...
+                        this.responses.write(response as StreamingHypothesis);
+                    }
+                }
+            });
+            this.doSendLoop(connection, this.requests);
+        });
+    }
+
+    private doSendLoop(connection: any, buffer: PassThrough): void {
+        if (connection.connected) {
```
amikofalvy: Is connection passed by reference?
Will the value of connected change in response to anything?

src/streaming-client.ts
amikofalvy: I liked what alex was saying about including example apps in our repository. The examples won't be packaged into the distributions, but they can be viewed/forked on github.

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+            let statusCode = null;
+            sut.on('httpResponse', code => {
+                statusCode = code;
+            });
+
+            // Act
+            mockClient.emit('httpResponse', {statusCode: 401});
+
+            // Assert
+            expect(statusCode).toBe(401);
```
amikofalvy: should we also assert that .end() is called?

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+            let expectedError = new Error("fake error")
+            sut.on('connectFailed', error => {
+                connectionError = error;
+            });
+            
+            // Act
+            mockClient.emit('connectFailed', expectedError);
+
+            // Assert
+            expect(connectionError).toBe(expectedError);
```
amikofalvy: ditto on asserting .end() was called.

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+            let connectionError = null;
+            const expectedError = new Error("fake connection error");
+            let mockConnection = new WebSocketConnection();
+            sut.on('error', error => {
+                connectionError = error;
+            });
+
+            // Act
+            mockClient.emit('connect', mockConnection);
+            mockConnection.emit('error', expectedError);
```
amikofalvy: Good test coverage.

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+            let mockConnection = new WebSocketConnection();
+            sut.on('connect', response => {
+                jobId = response.id;
+            });
+
+            // Act
+            mockClient.emit('connect', mockConnection);
+            mockConnection.emit('message', 
+                {
+                    type: 'utf8',
```
amikofalvy: Depending on your answer to the utf8 question above we might need another test for the non utf8 condition

test/streaming-client.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        });
+
+        it('emits connected event on connected message from server', () => {
+            // Setup
+            const res = sut.start();
+            let jobId = null;
+            const expectedJobId = "1";
+            let mockConnection = new WebSocketConnection();
+            sut.on('connect', response => {
+                jobId = response.id;
```
amikofalvy: do we get any other information in response?

We might as well validate all of the properties on response, not just id

amikofalvy: I see that you also pass `type`

src/streaming-client.ts
kbridbur: I make this call to avoid items which are gauranteed to throw if we try to cast them as Hypotheses. I could log something on non utf8 messages but I don't think we should throw an error

src/streaming-client.ts
kbridbur: Yes, connected will change when the connection is closed or errors. So in those cases we will stop sending data

test/streaming-client.spec.ts
kbridbur: :+1:

test/streaming-client.spec.ts
kbridbur: I don't think we need to test it as I don't think we should do anything in the case of non utf8 data

test/streaming-client.spec.ts
kbridbur: We get no other information. The connected response only has a job id right now

src/streaming-client.ts
amikofalvy: I was mostly curious what non utf8 messages we were being sent.

src/streaming-client.ts
amikofalvy: 👍 
> In Javascript objects and arrays follows pass by reference. Primitives are pass by value

test/streaming-client.spec.ts
amikofalvy: We can test for the "do nothing" behavior.

test/streaming-client.spec.ts
amikofalvy: In this test you are specifying the properties that are returned: you pass type and jobId

PR #9, opened 2019-06-06
AISDK-50: Upgrade axios to fix security vulnerability

kbridbur:


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "revai-node-sdk",
-  "version": "1.1.0",
+  "version": "2.0.0",
```
amikofalvy: You may need to re-run npm install. It looks like the version number increase didn't get added to package-lock.json

amikofalvy:
The CI tests are also failing. Please fix.

package-lock.json
kbridbur: Good catch

PR #10, opened 2019-06-10
AISDK-46: Add captions and file output to the node SDK

kbridbur:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "revai-node-sdk",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
amikofalvy: You haven't released the axios package fix yet? Why is this shown as a version increase?

amikofalvy: If you haven't released the axios fix, then you can release these two tickets together when we are ready to ship.

package.json
```diff
@@ -32,7 +32,7 @@
   },
   "homepage": "https://github.com/revdotcom/revai-node-sdk#readme",
   "dependencies": {
-    "axios": "^0.18.0",
+    "axios": "^0.19.0",
```
amikofalvy: ?

src/api-client.ts
```diff
@@ -167,11 +173,16 @@ export class RevAiApiClient {
...
-    async getTranscriptText(id: string): Promise<string> {
+    async getTranscriptText(id: string, fileName = null, fileLocation = null): Promise<string> {
         try {
             const response = await axios.get(`/jobs/${id}/transcript`, {
                 headers: { 'Accept': 'text/plain' }
             });
+
+            if (fileName) {
+                const fullPath = path.join(fileLocation || '', `${fileName}.txt`);
+                await this.writeOutputToFile(fullPath, response.data);
```
amikofalvy: This convenience feature wasn't a part of the Acceptance Criteria for this issue. Are we sure we want to add this feature? We will have to support it as we continue development on the SDK.

src/api-client.ts
```diff
@@ -167,11 +173,16 @@ export class RevAiApiClient {
         }
     }
 
-    async getTranscriptText(id: string): Promise<string> {
+    async getTranscriptText(id: string, fileName = null, fileLocation = null): Promise<string> {
```
amikofalvy: I think two parameters is a bit much to define the storage location for a file.
Can we combine the two into filePath? I think this will reduce confusion since the user doesn't have to worry about of the fileLocation must end in `/` or not.

amikofalvy: Have you thought about how the path must be specified? For example, can a user pass a relative path such as '~/user/username/myfile`?

src/api-client.ts
```diff
@@ -185,4 +196,36 @@ export class RevAiApiClient {
...
     }
+
+    async getCaptions(id: string, fileName = null, fileLocation = null): Promise<string> {
+        try {
+            const response = await axios.get(`/jobs/${id}/captions`, {
+                headers: { 'Accept': 'application/x-subrip' }
+            });
+
+            if (fileName) {
+                const fullPath = path.join(fileLocation || '', `${fileName}.srt`);
```
amikofalvy: Yes, I think it's premature to add save to file feature since we did not discuss it yet.

src/api-client.ts
```diff
@@ -185,4 +196,36 @@ export class RevAiApiClient {
...
+                case 409:
+                    throw new InvalidStateError(error);
+                default:
+                    throw error;
+            }
+        }
+    }
+
+    private async writeOutputToFile(filepath: string, data: string): Promise<void> {
+        await fs.writeFile(filepath, data, (err) => {
```
amikofalvy: I think mixing callbacks and async await reads poorly. Best practice when using await is to surround with `try {} catch {}`.

amikofalvy: Although, this is entirely dependent on writeFile returning a promise.
If it does return a promise, why do you have to manually throw the error when the exception should already be thrown?

package.json
kbridbur: I'm not sure what's going on with this. Develop definitely already has these changes. I think it may be because I changed the default branch on github

src/api-client.ts
kbridbur: I think we should add it. It's a very nice feature to have. I meant to add it as an AC

src/api-client.ts
```diff
@@ -149,10 +152,17 @@ export class RevAiApiClient {
     }
 
     async getTranscriptObject(id: string): Promise<RevAiApiTranscript> {
+        const stream = await this.getTranscriptObjectStream(id);
+        return stream.read();
```
amikofalvy: I think these changes break this function.

When I print the results to console:
```
<Buffer 7b 22 6d 6f 6e 6f 6c 6f 67 75 65 73 22 3a 5b 7b 22 73 70 65 61 6b 65 72 22 3a 30 2c 22 65 6c 65 6d 65 6e 74 73 22 3a 5b 7b 22 74 79 70 65 22 3a 22 74 ... >
```

And I can't access any of the properties. The types still show up in the IDE because we are casting raw bytes to our own type, but the returned object is invalid.

src/api-client.ts
```diff
@@ -149,10 +152,17 @@ export class RevAiApiClient {
...
+    }
+
+    async getTranscriptObjectStream(id: string): Promise<Readable> {
```
amikofalvy: Looks good to me. Let's discuss the longer term plan for defining output at the friday meeting. Example that we discussed earlier: should we have a singular function `client.getOutput(jobId, type, options)`?

src/api-client.ts
kbridbur: :( I thought I had fixed this. I'll look more into it

README.md
```diff
@@ -97,3 +103,19 @@ var transcriptObject = await client.getTranscriptObject(job.id);
...
+
+Any of these outputs can we retrieved as a stream for easy file writing:
+
+```javascript
+await client.getTranscriptTextStream(job.id);
+
+await client.getTranscriptObjectStream(job.id);
+```
+
+### Captions
```
amikofalvy: minor - what do you think about "Getting Caption Output"?

src/api-client.ts
```diff
@@ -17,6 +19,9 @@ export class RevAiApiClient {
     accessToken: string;
     version: string;
     instance: AxiosInstance;
+    private TEXT_OUTPUT = 'text/plain';
+    private JSON_OUTPUT = 'application/vnd.rev.transcript.v1.0+json';
+    private SRT_OUTPUT = 'application/x-subrip';
```
amikofalvy: 1. minor -  are these considered content-types? Should the name reflect this?

2. These should either be const at file scope of static instance members.

amikofalvy: Actually, these should be string enums https://www.typescriptlang.org/docs/handbook/enums.html
```
enum ContentType {
Text = 'text/plain',
Json = 'application/vnd.rev.transcript.v1.0+json',
Srt = 'application/x-subrip'
}
```

src/api-client.ts
```diff
@@ -149,29 +155,33 @@ export class RevAiApiClient {
...
+
+    async getTranscriptTextStream(id: string): Promise<Readable> {
+        return await this.getOutputsHelper(id, 'stream', this.TEXT_OUTPUT);
+    }
+
+    async getCaptions(id: string): Promise<Readable> {
+        return await this.getOutputsHelper(id, 'stream', this.SRT_OUTPUT);
+    }
+
+    private async getOutputsHelper(id: string, type: any, format: string): Promise<any> {
```
amikofalvy: `type` can also be declared as a string enum

amikofalvy: Looking at this some more: If you use this all purpose output helper, you should specify the return type as `Promise<Readable|string|RevAiApiTranscript>`. The `Promise<Readable|string|RevAiApiTranscript>` return type is ugly, but more well defined than `Promise<any>`

I believe, if you make the above suggestions you will be forced to cast the output of `getOutputsHelper` to one of the three possibilities in each of the convenience wrappers: getCaptions, getTranscriptText, etc.

I do worry about future sdk developers misusing the `getOutputsHelper` function and passing in incompatible types and formats.

amikofalvy: What do you think?

test/api-client.spec.ts
```diff
@@ -335,23 +337,22 @@ describe('api-client', () => {
...
             const resp = { data: jobDetails };
             mockedAxios.post.mockResolvedValue(resp);
 
             const job = await sut.submitJobLocalFile(filename);
 
             const expectedPayload = expect.objectContaining({
                 '_boundary': expect.anything(),
-                '_streams': expect.arrayContaining([expect.stringContaining('Content-Type: audio/mpeg'),
-                    expect.stringContaining('Content-Disposition: form-data; name=\"media\"; filename=\"test.mp3\"')])
+                '_streams': expect.anything()
```
amikofalvy: Did you remove assertions for this test?

test/api-client.spec.ts
```diff
@@ -361,8 +362,7 @@ describe('api-client', () => {
             };
             const expectedPayload = expect.objectContaining({
                 '_boundary': expect.anything(),
-                '_streams': expect.arrayContaining([expect.stringContaining('Content-Type: audio/mpeg'),
-                    expect.stringContaining('Content-Disposition: form-data; name="media"; filename="test.mp3"'),
+                '_streams': expect.arrayContaining([expect.anything(), expect.anything(),
```
amikofalvy: What's wrong with these assertions?

test/api-client.spec.ts
kbridbur: Good reminder. I no longer need to remove these. Those assertions no longer worked when the 'fs' import was mocked. I will add them back

test/api-client.spec.ts
```diff
@@ -435,15 +432,92 @@ describe('api-client', () => {
...
+            const resp = { data: expectedTranscript };
+            mockedAxios.get.mockResolvedValue(resp);
+
+            const transcript = await sut.getTranscriptObject(jobId);
+
+            expect(transcript).toEqual(expectedTranscript);
+        });
+
+        it('handles when api returns unauthorized', async () => {
+            const fakeError = setupFakeApiError(401, "Unauthorized");
```
amikofalvy: are we repeating these tests for all new endpoint related functions?

Maybe we should extract component which is a wrapper around axios and error handling. Then we can test the wrapper for proper error handling and the api-client only needs to assert that the wrapper was called with the correct params.

amikofalvy: The more I look at this file the more I think we need to extract the axios wrapper. It will greatly simplify test writing.

src/api-client.ts
kbridbur: I didn't realize this. I think that makes sense

kbridbur: The method is marked as private so they shouldn't be doing that. I don't think we should plan for customers using the SDK incorrectly and I also don't think we will run into any problems if they pass incorerct combinations. They'll just get errors from the api

test/api-client.spec.ts
kbridbur: Talked about this offline. It has been added as an AC for AISDK-49

README.md
```diff
@@ -97,3 +103,19 @@ var transcriptObject = await client.getTranscriptObject(job.id);
 ```
 
 The text output is a string containing just the text of your transcript. The object form of the transcript contains all the information outlined in the response of the [Get Transcript](https://www.rev.ai/docs#operation/GetTranscriptById) endpoint when using the json response schema.
+
+Any of these outputs can we retrieved as a stream for easy file writing:
+
+```javascript
+await client.getTranscriptTextStream(job.id);
```
amikofalvy: can we add a variable to show that these functions have a non void return type.
```suggestion
const textStream = await client.getTranscriptTextStream(job.id);
```

README.md
```diff
@@ -97,3 +103,19 @@ var transcriptObject = await client.getTranscriptObject(job.id);
...
+
+await client.getTranscriptObjectStream(job.id);
+```
+
+### Getting captions output
+
+Another way to retrieve your file is captions output. We currently only support .srt output. A readable stream of your srt output can be retrieved as such:
+
+```javascript
+await client.getCaptions(job.id);
```
amikofalvy: ditto

src/api-client.ts
amikofalvy: re:
> I do worry about future sdk developers misusing the getOutputsHelper function and passing in incompatible types and formats.

Here I am referring to future Rev developers, not to customer developers. That the function is private does not prevent rev developers from making mistakes.

amikofalvy: You don’t have to address anything in this PR. Just keep this in mind for when we extract an api wrapper.

PR #11, opened 2019-06-19
AISDK-49: Correctly use axios instance

kbridbur:


build_release_package_json.js
```diff
@@ -7,8 +7,8 @@ try {
     const pkg = JSON.parse(origPackage);
     delete pkg.devDependencies;
     delete pkg.scripts;
-    pkg.main = 'index.js';
-    pkg.module = 'index.js';
+    pkg.main = 'src/index.js';
+    pkg.module = 'src/index.js';
```
amikofalvy: do you need to specify that the entry point has changed in package.json?

src/api-client.ts
```diff
@@ -20,87 +14,61 @@ const enum ContentTypes {
...
 
-export class RevAiApiClient {
+export class RevAiApiClient extends ApiRequestHandler {
     accessToken: string;
     version: string;
-    instance: AxiosInstance;
     constructor (accessToken: string, version = 'v1') {
+        super(`https://api.rev.ai/revspeech/${version}/`, {
+            'Authorization': `Bearer ${accessToken}`,
+            'User-Agent': `RevAi-NodeSDK/2.0.1`
```
amikofalvy: Can we read the version from package.json now that you've changed the publish location of index.js?

src/api-client.ts
```diff
@@ -20,87 +14,61 @@ const enum ContentTypes {
     SRT = 'application/x-subrip'
 }
 
-export class RevAiApiClient {
+export class RevAiApiClient extends ApiRequestHandler {
```
amikofalvy: Why did you choose inheritance over composition?

I'm thinking from a testing perspective. To create components that are unit testable, I would prefer composition.
We could have two components: 
1. `RevAiApiClient`
2. `ApiRequestHelper`

The tests for the api methods in RevAiApiClient only need to assert that the methods of ApiRequestHelper were called with the correct parameters.

The tests for ApiRequestHelper assert that axios methods were called with the correct parameters and that different response status codes are handled appropriately.

amikofalvy: In the present state of your PR RevAiApiClient tests are overextending themselves and testing for implementation logic of the ApiRequestHandler. I think you can avoid this by switching to use composition instead of inheritance and refactoring the tests just a bit.

build_release_package_json.js
kbridbur: Good catch

src/api-client.ts
```diff
@@ -20,87 +14,63 @@ const enum ContentTypes {
...
+
 export class RevAiApiClient {
     accessToken: string;
-    version: string;
-    instance: AxiosInstance;
+    apiHandler: ApiRequestHandler;
     constructor (accessToken: string, version = 'v1') {
+        this.apiHandler = new ApiRequestHandler(`https://api.rev.ai/revspeech/${version}/`, {
+            'Authorization': `Bearer ${accessToken}`,
+            'User-Agent': `RevAi-NodeSDK/${sdkVersion}`
```
amikofalvy: The User Agent header does not need to be defined here. We could define it within ApiRequestHandler then we wouldn't have to pass it through the constructor.


Same with the `https://api.rev.ai/revspeech/` part of `https://api.rev.ai/revspeech/${version}/`.

I think we'll only need to pass the version number.

I would refactor the constructor of ApiRequestHandler to `ApiRequestHandler(accessToken: string, version: string)`

src/api-client.ts
```diff
@@ -20,87 +14,63 @@ const enum ContentTypes {
...
-        } catch (error) {
-            switch (error.response.status) {
-                case 401:
-                    throw new RevAiApiError(error);
-                default:
-                    throw error;
-            }
-        }
+        return await this.apiHandler.makeApiRequest<RevAiAccount>(
+            'get',
```
amikofalvy: We don't need to put each parameter on a new line.

You can put them all on the same line and wrap only the parameters that overflow the line character limit

src/api-request-handler.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import {
+    InsufficientCreditsError,
+    InvalidParameterError,
+    InvalidStateError,
+    RevAiApiError
+} from './models/RevAiApiError';
+
+export type HttpMethodTypes = 'post' | 'get' | 'delete';
+export type AxiosResponseTypes = 'stream' | 'json' | 'text';
```
amikofalvy: Is this type already defined within Axios?

amikofalvy: I don't want to duplicate the definition.

src/api-request-handler.ts
```diff
@@ -0,0 +1,66 @@
...
+        responseType: AxiosResponseTypes,
+        params?: {}
+    ): Promise<Response> {
+        try {
+            const data = (method === 'get' || method === 'delete') ? undefined : params;
+            const response = await this.instance.request({
+                method: method,
+                url: url,
+                data: data,
+                headers: headers,
```
amikofalvy: Just making sure, does this override the headers or is this an additive operation?

test/api-client.spec.ts
```diff
@@ -31,100 +19,66 @@ describe('api-client', () => {
...
-            expect(mockedAxios.get).toBeCalledTimes(1);
-        });
-
         it('get account email and balance', async () => {
             const accountEmail = 'test@rev.com';
             const balanceSeconds = 300;
             const data = { email: accountEmail, balance_seconds: balanceSeconds};
-            const resp = { data: data };
-            mockedAxios.get.mockResolvedValue(resp);
+            var mockHandler = ApiRequestHandler.mock.instances[0];
```
amikofalvy:
```suggestion
            const mockHandler = ApiRequestHandler.mock.instances[0];
```

src/api-request-handler.ts
kbridbur: It is not unfortunately

src/api-request-handler.ts
kbridbur: It does not. They are additive :)

PR #12, opened 2019-07-02
QAT-430

Chanc3:
- Reorganized the test folder with unit and integration packages
- Created integration project and added single jest test


test/integration/package.json
```diff
@@ -0,0 +1,29 @@
...
+  },
+  "scripts": {
+    "test": "jest"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/revdotcom/revai-node-sdk.git"
+  },
+  "author": "Rev Ai",
+  "license": "MIT",
```
amikofalvy:
```suggestion
  "license": "UNLICENSED",
```

test/integration/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "bugs": {
```
amikofalvy: You can remove this prop

test/integration/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "url": "https://github.com/revdotcom/revai-node-sdk/issues"
+  },
+  "homepage": "https://github.com/revdotcom/revai-node-sdk#readme",
```
amikofalvy: You can put the link to the integration sub project
```suggestion
  "homepage": "https://github.com/revdotcom/revai-node-sdk/test/integration",
```

test/integration/src/client-util.js
```diff
@@ -0,0 +1,10 @@
+const configUtil = require('./config-util');
+const revai = require('revai-node-sdk');
+
+module.exports = {
+    getClient: function (apiKey) {
+        let client = new revai.RevAiApiClient(apiKey);
```
amikofalvy:
```suggestion
        const client = new revai.RevAiApiClient(apiKey);
```

test/integration/src/config-util.js
```diff
@@ -0,0 +1,13 @@
+require('dotenv').config();
```
amikofalvy: is dotenv used?

We should remove it from the package.json if unused.

test/integration/test/account.test.js
```diff
@@ -0,0 +1,9 @@
+const clientUtil = require('../src/client-util');
+const configUtil = require('../src/config-util');
```
amikofalvy: Minor - I would call these helpers. In my mind Utils are equivalent to simple functions that reformat strings/dates/etc

test/integration/test/account.test.js
```diff
@@ -0,0 +1,9 @@
...
+
+test('Should return email and balance', async () => {
+    let client = clientUtil.getClient(configUtil.getApiKey());
```
amikofalvy:
```suggestion
    const client = clientUtil.getClient(configUtil.getApiKey());
```

test/integration/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "devDependencies": {
+    "revai-node-sdk": "^2.0.1",
```
kbridbur: In order to run theses on the latest version of the SDK we will probably have to import the SDK from the dist folder.

test/integration/src/client-util.js
```diff
@@ -0,0 +1,10 @@
+const configUtil = require('./config-util');
+const revai = require('revai-node-sdk');
+
+module.exports = {
+    getClient: function (apiKey) {
```
amikofalvy: we could use the arrow syntax

```suggestion
    getClient: (apiKey) => {
```

amikofalvy: You can also get the api key directly. I don't see a need to parameterize with the api key.
```suggestion
    getClient: () =>  {
        const client = new revai.RevAiApiClient(configUtil.getApiKey());
    }
```

test/integration/src/config-util.js
Chanc3: I assume that these variables will be passed into the test from ci but in order to run them locally they need to be populated from somewhere. How would you suggest I deal with these locally?

test/integration/src/client-util.js
Chanc3: Do we want any future test to check that the correct error is returned for an invalid api key? If not then I can make the change you mentioned.

JenkinsFile
```diff
@@ -20,7 +20,10 @@ pipeline {
             steps {
                 sh """
                     npm run lint
+                    node build_release_package_json.js
                     npm run test
+                    node build_release_package_json.js
```
amikofalvy, line 25: Why do we have to build release package again?

package.json
```diff
@@ -5,7 +5,8 @@
   "main": "src/index.js",
   "scripts": {
     "build": "tsc",
-    "test": "jest",
+    "test": "jest -c ./jest.config.unit.js",
```
amikofalvy:
```suggestion
    "unit-test": "jest -c ./jest.config.unit.js",
```

package.json
```diff
@@ -5,7 +5,8 @@
...
+    "integration": "jest -c ./jest.config.integration.js",
```
amikofalvy:
```suggestion
    "integration-test": "jest -c ./jest.config.integration.js",
```

What do you think about the above two suggestions?

test/integration/test/account.test.js
```diff
@@ -0,0 +1,22 @@
...
+const configHelper = require('../src/config-helper');
+
+test('Can return email and balance', async () => {
+    const client = clientHelper.getClient(configHelper.getApiKey());
+    const account = await client.getAccount();
+    expect(account.email).toBe(configHelper.getUserEmail());
+    expect(account.balance_seconds).not.toBeNull();
+});
+
+test('Cannot authenticate with fake token', async () => {
```
amikofalvy:
```suggestion
test('Cannot authenticate with invalid token', async () => {
```

test/integration/test/captions.test.js
```diff
@@ -0,0 +1,28 @@
+const clientHelper = require('../src/client-helper');
+const configHelper = require('../src/config-helper');
+
+test('Can get srt captions', async (done) => {
+    const client = clientHelper.getClient(configHelper.getApiKey());
+    const jobList = await client.getListOfJobs();
+    const jobId = getTranscribedJobId(jobList);
+    expect(jobId).toBeDefined();
```
amikofalvy, line 8: This test would fail in an environment/account that has not previously submitted a job.

test/integration/test/captions.test.js
```diff
@@ -0,0 +1,28 @@
...
+    captionsStream.on('data', data => {
+        streamString += data.toString();
+    })
+    captionsStream.on('end', () => {
+        expect(streamString.length).toBeGreaterThan(0)
+        done();
+    })
+})
+
+function getTranscribedJobId(jobList) {
```
amikofalvy: This function definition seems to be repeated across multiple files. I would suggest creating either a helper.

amikofalvy: consider defining this in client helpers?

test/integration/test/list.test.js
```diff
@@ -0,0 +1,18 @@
+const clientHelper = require('../src/client-helper');
+const configHelper = require('../src/config-helper');
+const RevAiApiJob = require('../../../dist/src/models/RevAiApiJob')
+
+test('Can get list of jobs', async () => {
```
amikofalvy: I would cluster the `list.test.js` and `job.test.js` in the same file `job.test.js` since the url is similar `/jobs`. 

Some interactions use POST, others use GET or DELETE.

Some require url params

amikofalvy (approved):
I think we are nearly ready to ship this issue. I left a few comments if we need to discuss further, ping me on slack, otherwise. :shipit:

JenkinsFile
kbridbur: We use the distribution to be deployed for the integration tests. The api request handler gets the SDK version for the user agent string from package.json. If this file doesn't exist it errors

test/integration/test/captions.test.js
Chanc3: You are correct. The SDK is using the same user as the other integration tests and since those run so often we should be ok for now. I will file a follow up ticket to address this in every test that expects a completed job is present.

PR #14, opened 2019-07-15
AISDK-47: Add examples to the node SDK

menioa:


package.json
```diff
@@ -34,7 +34,7 @@
   "dependencies": {
     "axios": "^0.19.0",
     "form-data": "^2.3.3",
-    "websocket": "^1.0.28"
+    "websocket": "^1.0.29"
```
kbridbur: I think this happened when you installed this manually inside the project directory. This is fine because it's only a minor version change but generally be careful about this so we don't create breaking changes

examples/streaming-examples.ts
```diff
@@ -0,0 +1,34 @@
...
+const audioConfig = new AudioConfig(<YOUR-CONTENT-TYPE-PARAMS>);
+const token = <YOUR-ACCESS-TOKEN>;
+var client = new RevAiStreamingClient(token, audioConfig);
+
+/* Create your event responses */
+client.on('close', (code, reason) => {
+	console.log(`Connection closed, ${code}: ${reason}`);
+});
+client.on('httpResponse', code => {
+	console.log(code);
```
kbridbur: Now that this is a public example and people will likely run it before they read it closely. Lets add a little more detail to these log statements

examples/streaming-examples.ts
```diff
@@ -0,0 +1,34 @@
...
+})
+client.on('connect', connectionMessage => {
+	console.log(connectionMessage);
+})
+
+/* Begin streaming session */
+var stream = client.start();
+
+/* Establish file for streaming and streaming responses */
+var file = fs.createReadStream(<YOUR-AUDIO-FILE>);
```
kbridbur: We should include a default audio file for them. Just use the english_raw one

examples/async-examples.ts
```diff
@@ -0,0 +1,47 @@
...
+const token = <YOUR-ACCESS-TOKEN>;
+var client = new RevAiApiClient(token);
+
+/* Get account details */
+var account = await client.getAccount();
+console.log(`Account: {account.email}`);
+console.log(`Balance: {account.balance_seconds} seconds`);
+
+/* Submit job and receive job object */
+var job = await client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3");
```
kbridbur: Add some extra options to these submits just to show that we have them

examples/async-examples.ts
```diff
@@ -0,0 +1,47 @@
...
+var jobDetails = await client.getJobDetails(job.id);
+
+/* Get all transcription jobs */
+var listOfJobs = await client.getListOfJobs();
+
+/* Get at most # of jobs */
+var listOfJobs = await client.getListOfJobs(5);
+
+/* Get all jobs started after a certain job id */
+var listOfJobs = await client.getListOfJobs(undefined, job.id);
```
kbridbur: I think you can just call this once with all the possible arguments and have a comment saying what the defaults are

examples/async-examples.ts
```diff
@@ -0,0 +1,47 @@
...
+var textStream = await client.getTranscriptTextStream(job.id);
+
+/* Get transcript as object stream */
+var transcriptStream = await client.getTranscriptObjectStream(job.id);
+
+/* Get captions */
+var captionsStream = await client.getCaptions(job.id);
+
+/* Delete a job */
+await client.deleteJob(job.id);
```
kbridbur: This should be commented out by default so they can see the job on their recent jbos page

examples/async-examples.ts
```diff
@@ -0,0 +1,47 @@
...
+var listOfJobs = await client.getListOfJobs();
+
+/* Get at most # of jobs */
+var listOfJobs = await client.getListOfJobs(5);
+
+/* Get all jobs started after a certain job id */
+var listOfJobs = await client.getListOfJobs(undefined, job.id);
+
+/* Get transcript as plain text */
+var transcriptText = await client.getTranscriptText(job.id);
```
kbridbur: Lets just print one of these out so they have some final output

examples/streaming-examples.ts
```diff
@@ -0,0 +1,43 @@
...
+stream.on('data', data => {
+	console.log(data);
+});
+stream.on('end', function () {
+	console.log("End of Stream");
+});
+
+/* Stream the file */
+file.pipe(stream);
+
```
kbridbur: Let's add a commented out `stream.end()` at the end of the file with a comment explaining the use of the method

examples/streaming-examples.ts
```diff
@@ -0,0 +1,43 @@
+import {AudioConfig, RevAiStreamingClient} from 'revai-node-sdk';
+import * as fs from 'fs';
+
+/* Initialize your client with your audio configuration and access token */
+const audioConfig = new AudioConfig(
+	/* contentType */ "audio/raw", 
+	/* layout */      "interleaved", 
+	/* sample rate */ 16000,
+	/* format */      "S16LE",
+    /* channels */    1
```
kbridbur: Formatting looks weird. Not sure if it's just github though

examples/async-examples.ts
```diff
@@ -0,0 +1,53 @@
...
+console.log(`Status for job ${job.id} is ${jobDetails.status}`);
+
+/* Get # of jobs started after a certain job id
+ * client.getListOfJobs() will default to returning all jobs
+ */
+var listOfJobs = await client.getListOfJobs(5, job.id);
+
+/* Get transcript as plain text
+ * Transcripts can also be gotten as Object, Text Stream, or Object Stream via
+ * client.getTranscript{Object, TextStream, ObjectStream}() respectively
```
kbridbur: This is a bit confusing to read and I think also inaccurate? All the possible retrieval methods are 
`getTranscriptText`, `getTranscriptTextStream`, `getTranscriptObject`, `getTranscriptObjectStream`, `getCaptionsStream`

kbridbur: Instead I would just have all the possible methods below but commented out

examples/streaming-examples.ts
menioa: I think its a combination of tabs and spaces, files have been adjusted

examples/async-examples.ts
```diff
@@ -0,0 +1,53 @@
+import {RevAiApiClient, RevAiJobOptions} from 'revai-node-sdk';
```
amikofalvy: minor -
```suggestion
import { RevAiApiClient, RevAiJobOptions } from 'revai-node-sdk';
```

examples/async-examples.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/* Initialize your client with your revai access token */
+const token = <YOUR-ACCESS-TOKEN>;
```
amikofalvy:
```suggestion
const token = '<YOUR-ACCESS-TOKEN>';
```

examples/async-examples.ts
```diff
@@ -0,0 +1,53 @@
...
+var client = new RevAiApiClient(token);
+
+/* Get account details */
+var account = await client.getAccount();
+console.log(`Account: ${account.email}`);
+console.log(`Balance: ${account.balance_seconds} seconds`);
+
+/* Submit job and receive job object
+ * Media may be submitted from a local file or from a url by using:
```
amikofalvy: I find the wording in this comment confusing.

Also the placement of this comment is far from the function: `submitJobLocalFile`

examples/async-examples.ts
```diff
@@ -0,0 +1,53 @@
...
+ * client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3");
+ */
+var jobOptions = new RevAiJobOptions(
```
amikofalvy: Where are the job options used.

examples/async-examples.ts
```diff
@@ -0,0 +1,53 @@
...
+	metadata = "This is an example",
+	skip_diarization = true
+	);
+var job = client.submitJobLocalFile("./example.raw");
```
amikofalvy: @kbridbur I would like to make another case for `submitJobFromStream(stream)`.

`client.submitJobFromStream(fs.createReadStream('./example.raw'))'`

amikofalvy: @menioa 
Try running the example: I don't think it will work. Missing an `await`.
```suggestion
var job = await client.submitJobLocalFile("./example.raw");
```

examples/async-examples.ts
menioa: This options implentation is incorrect, and it has been addressed. They are to be used in `submitJobLocalFile`

examples/async-examples.ts
menioa: imports are being restructured as the examples will be `.js` rather than `.ts`

examples/async-examples.js
```diff
@@ -0,0 +1,59 @@
+(async () => {	
+	const revai = require('../../revai-node-sdk');
```
amikofalvy: indent is too large. Is this a tab or space? It looks like a double tab

amikofalvy: Convention is to scope variables to their limited use; however, with imports (and import variables) I usually see them being scoped to the file rather than being imported inline dynamically. Scoping to the file level would allow you use the import ('revai') in private helper functions defined outside the scope of the main anon async function. 


Additionally I think we should structure the examples like this
```
revai-node-sdk/
├── src/
├── test/
└── examples/
    ├── async_transcribe_local_media_file/
    │   ├── async_transcribe_local_media_file.js
    │   └── package.json
    ├── async_transcribe_media_from_url/
    │   ├── async_from_local_file.js
    │   └── package.json
    ├── streaming_transcribe_from_local_file/
    │   ├── streaming_transcribe_from_local_file.js
    │   └── package.json
    ├── streaming_transcribe_from_stream/
    │   ├── streaming_transcribe_from_stream.js
    │   └── package.json
    └── streaming_transcribe_from_live_microphone/
        ├── streaming_transcribe_from_live_microphone.js
        └── package.json
```


Then we can use the imports without using the relative path

```suggestion
	const revai = require('revai-node-sdk');
```


My reason for

examples/async-examples.js
```diff
@@ -0,0 +1,59 @@
...
+	/* Initialize your client with your revai access token */
+	const token = '<YOUR-ACCESS-TOKEN>';
+	var client = new revai.RevAiApiClient(token);
+
+	/* Get account details */
+	var account = await client.getAccount();
+	console.log(`Account: ${account.email}`);
+	console.log(`Balance: ${account.balance_seconds} seconds`);
+
+
```
amikofalvy: minor - extra space

examples/async-examples.js
```diff
@@ -0,0 +1,59 @@
...
+	var client = new revai.RevAiApiClient(token);
+
+	/* Get account details */
+	var account = await client.getAccount();
+	console.log(`Account: ${account.email}`);
+	console.log(`Balance: ${account.balance_seconds} seconds`);
+
+	
+	const jobOptions = {
+	    metadata: "This is an example",
```
amikofalvy: Let's use a more practical example.


```suggestion
	    metadata: "InternalOrderNumber=364",
```

examples/async-examples.js
```diff
@@ -0,0 +1,59 @@
...
+	    skip_diarizationh: true
```
amikofalvy: typo
```suggestion
	    skip_diarization: true
```

examples/async-examples.js
```diff
@@ -0,0 +1,59 @@
...
+	};
+
+	/* Media may be submitted from a local file or from a url by using:
+	 * client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3");
+	 */
+	var job = await client.submitJobLocalFile("./example.mp3", jobOptions);
```
amikofalvy: I think resources should be put into a common folder

examples/async-examples.js
```diff
@@ -0,0 +1,59 @@
...
+	/* Check the status of job */
+	var jobDetails = await client.getJobDetails(job.id);
+	console.log(`Status for job ${job.id} is ${jobDetails.status}`);
+
+	/* Get # of jobs started after a certain job id
+	 * client.getListOfJobs() will default to returning all jobs
+	 */
+	var listOfJobs = await client.getListOfJobs(5, job.id);
+
+	/* Get transcript as plain text
```
amikofalvy: The code directly below the comment does not get the transcript as plain text. It loops with a 5 second delay until the job is in a terminal state.

examples/async-examples.js
```diff
@@ -0,0 +1,59 @@
...
+	    // Waits 5 seconds between each check
+	    console.log("looping");
+	    await new Promise( resolve => setTimeout(resolve, 5000));
+	}
+
+	var transcriptText = await client.getTranscriptText(job.id);
+	// var transcriptTextStream = await client.getTranscriptTextStream(job.id);
+	// var transcriptObject = await client.getTranscriptObject(job.id);
+	// var transcriptObjectStream = await client.getTranscriptObjectStream(job.id);
+	// var captionsStream = await client.getCaptions(job.id);
```
amikofalvy: I think we should write these to file and console.log('Look in the directory for the captions file');

examples/file-streaming-examples.js
```diff
@@ -0,0 +1,45 @@
+const revai = require('../../revai-node-sdk');
+const fs = require('fs');
+
+/* Initialize your client with your audio configuration and access token */
+const audioConfig = new revai.AudioConfig(
+    /* contentType */ "audio/x-raw",
```
amikofalvy: @kbridbur Looking at this usage without typescript, it makes me eager to change this constructor to accept an object parameter.

```
// signature
new AudioConfig(options)

// usage
new AudioConfig({
content_type: 'audio/x-raw',
layout: 'interleaved',
sample_rate: 16000,
format: 'S16LE',
channels: 1
})


```

examples/file-streaming-examples.js
```diff
@@ -0,0 +1,45 @@
...
+    console.log(`Connection failed with error: ${error}`);
+})
+client.on('connect', connectionMessage => {
+    console.log(`Connected with message: ${connectionMessage}`);
+})
+
+/* Begin streaming session */
+var stream = client.start();
+
+/* Establish file for streaming and streaming responses */
```
amikofalvy:
```suggestion
/* Read file from disk */
```

examples/file-streaming-examples.js
```diff
@@ -0,0 +1,45 @@
...
+});
+stream.on('end', function () {
+    console.log("End of Stream");
+});
+
+/* Stream the file */
+file.pipe(stream);
+ 
+/* Ends the streaming session */
+// stream.end();
```
amikofalvy: code is commented out.

examples/async-examples.js
amikofalvy: For Kyle: I asked Albert to stop working on the microphone nodejs streaming example because nodejs doesn't have a nice microphone package. I think it would be more valuable to create a client/server app that is basically simplified homepage streaming demo.

examples/async-examples.js
```diff
@@ -0,0 +1,59 @@
+(async () => {	
+	const revai = require('../../revai-node-sdk');
+
+	/* Initialize your client with your revai access token */
+	const token = '<YOUR-ACCESS-TOKEN>';
```
amikofalvy: +1 for using a config file both for development and for publishing to prevent us from accidentally committing access_tokens to feature branches while developing

examples/async-examples.js
amikofalvy: Another comment; can we update the readme.md to let customers know that we have examples.

amikofalvy: Also, we should write down which version of node we support. Or is this something we can add as a travis ci type badge to the readme?

examples/async-examples.js
JDongian: Do you have a linter to find this for you?

examples/async-examples.js
JDongian: One idea is to have a public sample access token that actually works, but is heavily throttled. I don't support the idea.

examples/async-examples.js
JDongian: I think if you're going to have magic numbers, make sure they are well-known (people could get confused and think they are meaningful). e.g. `42`, `9001` are pretty good. I think best is to avoid any possible confusion.

examples/file-streaming-examples.js
JDongian: For my own curiosity, what are the primary arguments in favor of building an object like this?

examples/async-examples.js
kbridbur: We do have a linter. @menioa run `npm run lint`

examples/file-streaming-examples.js
kbridbur: Josh I'm not quite sure I understand your question.

Andrew I've considered this and think it would be a reasonable change. My main issue being that it is a breaking change

examples/file-streaming-examples.js
kbridbur: This code should be commented out. In theory `.end()` should not be necessary to call when streaming from a file

examples/async-examples.js
kbridbur: Would a client server app with a streaming demo not require getting audio from microphone anyways? Can you clarify the specs of the app example? 

It should say in the readme, please add it if it doesn't. We support node 8 and 10.

examples/async-examples.js
kbridbur: Using a config file is a very good idea. I'm not sure where we would expect them to put the file though. I'm sure other node packages do this so I'll look into what they do.

I definitely don't support the throttled public token either

examples/async-examples.js
menioa: I believe by using a client server app would allow microphone access to come via the browser rather than directly from node.

examples/async_get_job_details/async_get_job_details.js
```diff
@@ -0,0 +1,29 @@
...
+(async () => {  
+    /* Initialize your client with your revai access token */
+    var client = new revai.RevAiApiClient(token);
+
+    /* Get account details */
+    var account = await client.getAccount();
+    console.log(`Account: ${account.email}`);
+    console.log(`Balance: ${account.balance_seconds} seconds`);
+
+    const jobOptions = {
```
kbridbur: We should either show all job options or none. Since this example is about job details I think we should just show none

examples/async_get_job_details/async_get_job_details.js
```diff
@@ -0,0 +1,29 @@
...
+
+    var filenames = ["example.mp3", "example.mp3"];
+    filenames.map(async filename => await client.submitJobLocalFile(`..\\resources\\${filename}`, jobOptions));
+
+    /* Retrieves a list of jobs. */
+    var jobList = await client.getListOfJobs(5, null);
+    console.log(jobList);
+
+    /* Delete a job */
+    // await client.deleteJob(job.id);
```
kbridbur: This line would error if they were to uncomment it. It also seems somewhat irrelevant to this example

examples/async_transcribe_local_media_file/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,53 @@
...
+    /* Initialize your client with your revai access token */
+    var client = new revai.RevAiApiClient(token);
+
+    /* Get account details */
+    var account = await client.getAccount();
+    console.log(`Account: ${account.email}`);
+    console.log(`Balance: ${account.balance_seconds} seconds`);
+
+    const jobOptions = {
+        metadata: "InternalOrderNumber=123456789",
```
kbridbur: Same as above. We should show examples for every property

examples/async_transcribe_local_media_file/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,53 @@
...
+        skip_diarization: true
+    };
+
+    /* Media may be submitted from a local file or from a url by using:
```
kbridbur: This example is specific to local file submit. This comment should be changed

examples/async_transcribe_local_media_file/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,53 @@
...
+     * client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3");
+     */
+    var job = await client.submitJobLocalFile("../resources/example.mp3", jobOptions);
+
+    console.log(`Job Id: ${job.id}`);
+    console.log(`Status: ${job.status}`);
+    console.log(`Created On: ${job.created_on}`);
+
+    // Waits 5 seconds between each status check to see if job is complete
```
kbridbur: Do you think it makes sense to add a comment explaining why we wait?

examples/async_transcribe_local_media_file/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,53 @@
...
+     * Transcripts can also be gotten as Object, Text Stream, Object Stream,
+     * or as captions
+     */
+    var transcriptText = await client.getTranscriptText(job.id);
+    // var transcriptTextStream = await client.getTranscriptTextStream(job.id);
+    // var transcriptObject = await client.getTranscriptObject(job.id);
+    // var transcriptObjectStream = await client.getTranscriptObjectStream(job.id);
+    // var captionsStream = await client.getCaptions(job.id);
+
+    fs.writeFile("./transcript.txt", transcriptText, (err) => {
```
kbridbur: If we're going to write output files we should add an "outputs" directory

examples/async_transcribe_local_media_file/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,53 @@
...
+        if (err) throw err;
+        console.log("Success! Check you directory for the transcript.")
```
kbridbur: Typo `your`. We should also say the directory name

examples/async_transcribe_media_from_url/async_transcribe_media_from_url.js
```diff
@@ -0,0 +1,53 @@
...
+
+    const jobOptions = {
+        metadata: "InternalOrderNumber=123456789",
+        skip_diarization: true
+    };
+
+    /* Media may be submitted from a local file or from a url by using:
+     * client.submitJobLocalFile("../resources/example.mp3", jobOptions);
+     */
+    var job = await client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3");
```
kbridbur: Why don't we use the options here

examples/async_transcribe_media_from_url/async_transcribe_media_from_url.js
```diff
@@ -0,0 +1,53 @@
...
+    console.log(`Account: ${account.email}`);
+    console.log(`Balance: ${account.balance_seconds} seconds`);
+
+    const jobOptions = {
+        metadata: "InternalOrderNumber=123456789",
+        skip_diarization: true
+    };
+
+    /* Media may be submitted from a local file or from a url by using:
+     * client.submitJobLocalFile("../resources/example.mp3", jobOptions);
```
kbridbur: This comment is also unneeded

examples/async_transcribe_media_from_url/async_transcribe_media_from_url.js
```diff
@@ -0,0 +1,53 @@
...
+     */
+    var job = await client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3");
+
+    console.log(`Job Id: ${job.id}`);
+    console.log(`Status: ${job.status}`);
+    console.log(`Created On: ${job.created_on}`);
+
+    // Waits 5 seconds between each status check to see if job is complete
```
kbridbur: Same as the other example

examples/streaming_transcribe_from_local_file/file-streaming-examples.js
```diff
@@ -0,0 +1,47 @@
+const revai = require('revai-node-sdk');
+const fs = require('fs');
+const token = require('../config/config.json').access_token;
+
+/* Initialize your client with your audio configuration and access token */
+const audioConfig = new revai.AudioConfig(
```
kbridbur: Nice job fixing the formatting

examples/streaming_transcribe_from_local_file/file-streaming-examples.js
```diff
@@ -0,0 +1,47 @@
...
+
+stream.on('data', data => {
+    console.log(data);
+});
+stream.on('end', function () {
+    console.log("End of Stream");
+});
+
+/* Stream the file */
+file.pipe(stream);
```
kbridbur: This will have to change a little once AISDK-54 is merged

.gitignore
```diff
@@ -60,3 +60,9 @@ typings/
 
 # next.js build output
 .next
+
+# example config file
+examples/config/config.json
+
+# Example transcript output
+transcript.txt
```
kbridbur: This would also be easier if we added an outputs folder :) we could just ignore everything inside it

examples/async_transcribe_media_from_url/async_transcribe_media_from_url.js
menioa: It was a mistake, resolved

examples/async_transcribe_local_media_file/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,62 @@
...
+    };
+
+    /* Media may be submitted from a local file */
+    var job = await client.submitJobLocalFile("../resources/example.mp3", jobOptions);
+
+    console.log(`Job Id: ${job.id}`);
+    console.log(`Status: ${job.status}`);
+    console.log(`Created On: ${job.created_on}`);
+
+    /* Waits 5 seconds between each status check to see if job is complete.
```
kennygea: are comments supposed to start on this line?

or are they supposed to be like 
```
/*
* Media may be..
*/
```
I'm just as unfamiliar with typescript as you, so question for @kbridbur or @amikofalvy

examples/async_transcribe_media_from_url/async_transcribe_media_from_url.js
```diff
@@ -0,0 +1,60 @@
...
+
+    /* Get account details */
+    var account = await client.getAccount();
+    console.log(`Account: ${account.email}`);
+    console.log(`Balance: ${account.balance_seconds} seconds`);
+
+    const jobOptions = {
+        metadata: "InternalOrderNumber=123456789",
+        callback_url: "https://jsonplaceholder.typicode.com/posts",
+        skip_diarization: true,
```
kennygea: should we add `skip_punctuation: false` here as well?

menioa: skip_punctuation is not in the RevAiJobOptions interface, so I think no

examples/async_transcribe_media_from_url/async_transcribe_media_from_url.js
```diff
@@ -0,0 +1,60 @@
...
+    };
+
+    /* Media may be submitted from a url */
+    var job = await client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3", jobOptions);
+
+    console.log(`Job Id: ${job.id}`);
+    console.log(`Status: ${job.status}`);
+    console.log(`Created On: ${job.created_on}`);
+
+    // Waits 5 seconds between each status check to see if job is complete
```
kennygea: should we add a `note: polling for job status is not recommended in a non-testing environment. Use the callback_url option [link to documentation] to receive the response asynchronously on job completion`?

examples/streaming_transcribe_from_local_file/streaming_transcribe_from_local_file.js
```diff
@@ -0,0 +1,47 @@
+const revai = require('revai-node-sdk');
+const fs = require('fs');
+const token = require('../config/config.json').access_token;
+
+/* Initialize your client with your audio configuration and access token */
```
kennygea: I wonder if it will be confusing for customers to stream non-raw audio file from this example, i.e they have to creat a new client with different `AudioConfig` parameters which is not obvious how to do so from this example. 

We have both local and media url async examples, should we add a non-raw (like a .wav) and stream? Or do you think that's overkill @kbridbur

examples/file-streaming-examples.js
kennygea: Josh's question I think (let me know if I'm off the mark) is what's the difference between declaring

````
func(var a, var b, var c)
{
}
```

vs

```
func(var obj)
{
}
```
where we have a container class `obj` which contains `var a, b, c` (this is for a constructor, but the same principals apply I think)

Preface by saying I do think it's case specific, but my arguments for usually grouping method inputs into one object is two fold. 

1. It makes it more clear what you're passing in (note you have to comment it w/o typescript)
In C# and typescript, we have named parameters, so this is alleviated somewhat, but not always which leads to 
2. that usually it's easier to track the stack call if/when you pass `obj` or it's properties around. This is more of an opinion of mine but explanation below.

There are 5 properties passed into the constructor of `AudioConfig`, and thankfully they are all related, so it's not too bad.

But say there was a 6th property like, `bool ShouldBufferAudio`. This isnt really related to the first 5, other than it's a part of `AudioConfig`. For this constructor, it's not a big deal since it's usually just setting properties, but if it was a method and the method is doing a wide variety of operations on the input parameters, it can get confusing real quick to trace what property came from where. By grouping them into logical groups, it makes it (once again, imo) easier to read.

JDongian: Makes sense, thanks. Basically the main use case of PODs.

examples/async_transcribe_local_media_file/async_transcribe_local_media_file.js
kbridbur: They should go on the next line. Single line comments begin with `//`

examples/streaming_transcribe_from_local_file/streaming_transcribe_from_local_file.js
kbridbur: I think it's a bit overkill. We should link the streaming docs in the readme though

README.md
```diff
@@ -4,8 +4,11 @@
 
 ## Documentation
 
-See the [API docs](https://www.rev.ai/docs) for more information about the API and
-more examples.
+See the [API docs](https://www.rev.ai/docs) for more information about the API.
+
+## Examples
+
+Examples are in the [`examples/`](https://github.com/revdotcom/revai-node-sdk/tree/develop/samples) directory
```
kbridbur: Minor: "doc speak" this should be worded as `can be found in`

examples/README.md
```diff
@@ -0,0 +1,29 @@
+# Rev.ai Node SDK Examples
+### Setup:
+From the `examples/` directory run `npm install`
+In the `config/` directory,  copy `sample.config.json` as `config.json`
```
kbridbur: Opinion: this would be more clear as `In config/sample.config.json, insert your access token where specified and rename the file to config.json`

examples/async_get_job_details.js
```diff
@@ -0,0 +1,21 @@
...
+(async () => {  
+    // Initialize your client with your revai access token
+    var client = new revai.RevAiApiClient(token);
+
+    // Get account details
+    var account = await client.getAccount();
+    console.log(`Account: ${account.email}`);
+    console.log(`Balance: ${account.balance_seconds} seconds`);
+
+    var filenames = ["example.mp3", "example.mp3"];
```
kbridbur: This seems unnecessary since they're both the same. I would see the point if it was easily replaceable but the way this is set up requires that the filename here is relative to the resources folder. I would just get rid of this array and write the filename in the submit

examples/async_transcribe_media_from_url/async_transcribe_media_from_url.js
kbridbur: This shows as outdated but has not been addressed @menioa please do this

examples/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,64 @@
...
+    // var transcriptObject = await client.getTranscriptObject(job.id);
+    // var transcriptObjectStream = await client.getTranscriptObjectStream(job.id);
+    // var captionsStream = await client.getCaptions(job.id);
+
+    fs.writeFile("../outputs/async_file_transcript.txt", transcriptText, (err) => {
+        if (err) throw err;
+        console.log("Success! Check the examples/outputs/ directory for the transcript.")
+    });
+
+    // Delete a job
```
kbridbur: We should note that this will remove all info stored about the job from our servers

examples/async_get_job_details.js
```diff
@@ -0,0 +1,21 @@
...
+    filenames.map(async filename => await client.submitJobLocalFile(`..\\resources\\${filename}`);
+
+    // Retrieves a list of jobs.
+    var jobList = await client.getListOfJobs(5, null);
```
kbridbur: Please make a note about what the second parameter does

examples/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,64 @@
...
+        metadata: "InternalOrderNumber=123456789",
+        callback_url: "https://jsonplaceholder.typicode.com/posts",
+        skip_diarization: true,
+        custom_vocabularies: [{
+            phrases: [
+                "add",
+                "custom",
+                "vocabularies",
+                "here"
+                ]
```
kbridbur: Minor formatting. Both this line and the next one should be dedented

examples/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,64 @@
...
+    var client = new revai.RevAiApiClient(token);
+
+    // Get account details
+    var account = await client.getAccount();
+    console.log(`Account: ${account.email}`);
+    console.log(`Balance: ${account.balance_seconds} seconds`);
+
+    const jobOptions = {
+        metadata: "InternalOrderNumber=123456789",
+        callback_url: "https://jsonplaceholder.typicode.com/posts",
```
kbridbur: Have we tested this and made sure it works? Does the job fail?

examples/async_transcribe_media_from_url.js
```diff
@@ -0,0 +1,64 @@
...
+        metadata: "InternalOrderNumber=123456789",
+        callback_url: "https://jsonplaceholder.typicode.com/posts",
+        skip_diarization: true,
+        custom_vocabularies: [{
+            phrases: [
+                "add",
+                "custom",
+                "vocabularies",
+                "here"
+                ]
```
kbridbur: Same here

examples/async_transcribe_media_from_url.js
```diff
@@ -0,0 +1,64 @@
...
+
+    // Get account details
+    var account = await client.getAccount();
+    console.log(`Account: ${account.email}`);
+    console.log(`Balance: ${account.balance_seconds} seconds`);
+
+    const jobOptions = {
+        metadata: "InternalOrderNumber=123456789",
+        callback_url: "https://jsonplaceholder.typicode.com/posts",
+        skip_diarization: true,
```
kbridbur: Lets set this to false. The final output without this doesn't show the full abilities of our speech to text and these outputs will likely be a first impression

examples/async_transcribe_local_media_file.js
```diff
@@ -0,0 +1,64 @@
...
+        skip_diarization: true,
```
kbridbur: Set to false

examples/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dependencies": {
+        "revai-node-sdk": "^2.0.1"
```
kbridbur: We should make this `"revai-node-sdk": "*"`. This way they will always get the latest version of the sdk

examples/streaming_transcribe_from_local_file.js
```diff
@@ -0,0 +1,47 @@
...
+    console.log(data);
+});
+stream.on('end', function () {
+    console.log("End of Stream");
+});
+
+// Stream the file
+file.pipe(stream);
+ 
+// Ends the streaming session
```
kbridbur: `Forcibly ends the streaming session`

package.json
```diff
@@ -34,6 +34,8 @@
   "dependencies": {
     "axios": "^0.19.0",
     "form-data": "^2.3.3",
+    "mic": "^2.1.2",
```
kbridbur: Remove these since we do not use them. These are also likely causing all the changes to package lock

examples/async_get_job_details.js
menioa: An array is used to demonstrate the getListOfJobs functionality. I think for this function to be properly represented it may be better to add another example file rather than submitting only one job.

package.json
kbridbur: Please address this comment

examples/async_get_job_details.js
```diff
@@ -0,0 +1,26 @@
...
+    var account = await client.getAccount();
+    console.log(`Account: ${account.email}`);
+    console.log(`Balance: ${account.balance_seconds} seconds`);
+
+    var filenames = [".\\resources\\example.mp3", ".\\resources\\example.mp3"];
+    filenames.map(async filename => await client.submitJobLocalFile(`${filename}`));
+
+    /**
+     * Retrieves a list of jobs in order from newest to oldest.
+     * The first parameter limits the number of jobs returned..
```
kbridbur:
```suggestion
     * The first parameter limits the number of jobs returned.
```

examples/async_get_job_details.js
```diff
@@ -0,0 +1,26 @@
...
+     * The second parameter allows for a job id to be set, indicating that only jobs
+     * submitted after that job will be returned
```
kbridbur:
```suggestion
     * submitted after that job will be returned.
```

examples/README.md
```diff
@@ -0,0 +1,31 @@
...
+View the [source code](https://github.com/revdotcom/revai-node-sdk/tree/develop/examples/async_transcribe_local_media_file.js).<br>
+**Run:**<br>
+`node async_transcribe_local_media_file.js`<br>
+The output can be found at `outputs/async_file_transcript.txt`
+
+### Transcription from a Media Url
+View the [source code](https://github.com/revdotcom/revai-node-sdk/tree/develop/examples/async_transcribe_media_from_url.js).<br>
+**Run:**<br>
+`node async_transcribe_media_from_url.js`<br>
+The output can be found at `examples/outputs/async_url_transcript.txt`
```
kbridbur: This path is inconsistent with the one above.

PR #15, opened 2019-07-22
AISDK-54: Allow client to send EOS without killing connection

kbridbur:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "revai-node-sdk",
-  "version": "2.0.1",
+  "version": "2.1.0",
```
kbridbur: I did a minor version update since this changes the functionality of the SDK a reasonable amount but doesn't break existing implementations

src/streaming-client.ts
```diff
@@ -94,12 +98,19 @@ export class RevAiStreamingClient extends EventEmitter {
...
+     * Signals to the api that you have finished sending data.
+     */
+    public finish(): void {
+        this.requests.end('EOS', 'utf8');
+    }
+
     /**
      * Ends the streaming connection and closes off the buffer returned from start()
      */
     public end(): void {
```
kennygea: what's the different between the `finish()` and `end()` methods?

In `finish` you do `this.requests.end('EOS', 'utf8')` but in `end`, you do just `this.request.end()` with no `EOS` or `utf8`

kbridbur: Finish signals to the server that the client is done giving input. End immediately cuts off the connection and stops everything

amikofalvy: Maybe we can combine the two so that there is only end which sends EOS and ends the connection . Why should we keep them separate?

kbridbur: The difference is that a user may not want to end the connection immediately.

If anything I think we should remove the end method since it technically breaks the contract we have (the user shouldn't be the one to end the connection)

src/streaming-client.ts
```diff
@@ -94,12 +98,19 @@ export class RevAiStreamingClient extends EventEmitter {
         return new BufferedDuplex(this.requests, this.responses, { readableObjectMode: true });
     }
 
+    /**
+     * Signals to the api that you have finished sending data.
+     */
+    public finish(): void {
```
kennygea: I would call this `end()` and rename the current end to `unsafeEnd()` and document that `unsafeEnd` immediately closes the connection and that no output will be returned.

kennygea: since `end` is the contract that our (non-existent) customers would be using to close their connections

kbridbur: I like this suggestion. It avoids a breaking change and makes sense as an implementation

kennygea (approved):
:shipit:

src/streaming-client.ts
```diff
@@ -117,6 +125,9 @@ export class RevAiStreamingClient extends EventEmitter {
             let value = buffer.read(buffer.readableLength);
             if (value !== null) {
                 connection.send(value);
+                if (value.includes('EOS') || value.includes(Buffer.from('EOS'))) {
+                    connection.sendUTF('EOS');
```
kennygea: this sends text type `EOS` now right?

kbridbur: Yes. sendUTF sends a text type message

PR #16, opened 2019-07-31
AISDK-56: Node SDK response object should support punctuation

Ymardini:


Ymardini:
Waiting for REVAI-436 to continue with this issue.

src/models/streaming/StreamingResponses.ts
```diff
@@ -11,10 +11,9 @@ export interface StreamingResponse {
  * Represents a hypothesis response from the rev.ai streaming service. Both final and partial.
  */
 export interface StreamingHypothesis extends StreamingResponse {
-    transcript?: string;
     ts?: number;
     end_ts?: number;
-    elements: Element[];
+    elements?: Element[];
```
kbridbur: I'm not sure this `?` is needed since `elements` can just be an empty array if it doesn't exist. Let's check once punctuation is on production

PR #18, opened 2019-08-07
AISDK-57: Fix security vulnerabilities

k-weng:
Vulnerabilities found using `npm audit` and fixed using `npm audit fix`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,22 +1,134 @@
 {
   "name": "revai-node-sdk",
-  "version": "2.0.1",
+  "version": "2.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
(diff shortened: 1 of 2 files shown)


JDongian:
Should we put this as a git hook or something? Or add it to the CI pipeline?

k-weng:
Yeah we could add a script to Travis CI. There's a [npm package](https://www.npmjs.com/package/audit-ci) we could use. @kbridbur

kbridbur:
I would rather not add it as an automatic step, it seems like something we should do manually. In this case it's fine since they're all dev dependencies but it seems like `npm audit fix` has no issue bumping major version numbers of dependencies which could theoretically cause a breaking change

kbridbur (approved):
:shipit: 

This PR looks good. Let's think a bit more about whether or not we want to automate dependency fixes

k-weng:
@kbridbur Didn't mention this in the PR description but `npm audit fix` won't actually fix any vulnerabilities that require major version bumping unless you give it the option to. In addition, the npm package that I reference only fails the build if `npm audit` finds any vulns.

kbridbur:
Ahh, I see. If that's the case then we should investigate adding a step like you guys are talking about to the CI pipeline. I'll file an issue

PR #19, opened 2019-08-10
AISDK-55: Catch SDK up to API in features

kbridbur:


test/api-client/api-client.captions.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { RevAiApiClient } from '../../src/api-client';
```
kbridbur, line 3: I also took this opportunity to clean up and split up our client tests :)

README.md
```diff
@@ -116,8 +116,8 @@ var transcriptStream = await client.getTranscriptObjectStream(job.id);
 
 ### Getting captions output
 
-Another way to retrieve your file is captions output. We currently only support .srt output. A readable stream of your srt output can be retrieved as such:
+Another way to retrieve your file is captions output. We support both .srt and .vtt outputs. A readable stream of your desired captions output can be retrieved as shown below. If your job was submitted as a multichannel file you will be required to provide the id of the channel you would like captioned.
```
amikofalvy: minor - the second sentence reads strange when it is passive.
```suggestion
Another way to retrieve your file is captions output. We support both .srt and .vtt outputs. See below for an example showing how you can get captions as a readable stream. If your job was submitted with multiple speaker channels you are required to provide the id of the channel you would like captioned.
```

README.md
```diff
@@ -116,8 +116,8 @@ var transcriptStream = await client.getTranscriptObjectStream(job.id);
...
 
 ```javascript
-var captionsStream = await client.getCaptions(job.id);
+var captionsStream = await client.getCaptions(job.id, CaptionTypes.SRT, null);
```
amikofalvy: I find the example confusing. As a customer, I would never provide null when there is a default value for this parameter.

kbridbur: I want to show that the parameter exists but in most cases it will be null and error if soemthing is provided. Suggestions?

src/api-client.ts
```diff
@@ -70,26 +70,30 @@ export class RevAiApiClient {
...
 
-    async getCaptions(id: string): Promise<Readable> {
+    async getCaptions(id: string, contentType?: CaptionTypes, channelId?: number): Promise<Readable> {
+        let url = `/jobs/${id}/captions`;
+        if (channelId) {
+            url += `?speaker_channel=${channelId}`;
+        }
         return await this.apiHandler.makeApiRequest<Readable>('get',
-            `/jobs/${id}/captions`, { 'Accept': ContentTypes.SRT }, 'stream');
+            url, { 'Accept': contentType || CaptionTypes.SRT }, 'stream');
```
amikofalvy, line 97: clever js logical or

test/api-client/api-client.captions.spec.ts
amikofalvy: I can't see the diffs as well when every line is moved to a new file, but I enjoy better separation of the tests

test/api-client/api-client.captions.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+import { ApiRequestHandler } from '../../src/api-request-handler';
+import { CaptionTypes } from '../../src/models/async/CaptionTypes';
+import { objectToStream } from '../testhelpers';
+const fs = require('fs');
```
amikofalvy: Why can't we use the import syntax here?
```suggestion
import * as fs from 'fs';
```

test/api-client/api-client.captions.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+                { 'Accept': contentType }, 'stream');
+            expect(mockHandler.makeApiRequest).toBeCalledTimes(1);
+        });
+
+        it('attaches channelId if given', async () => {
+            const mockHandler = ApiRequestHandler.mock.instances[0];
+            
+            await sut.getCaptions(jobId, null, 1);
+
+            expect(mockHandler.makeApiRequest).toBeCalledWith('get', `/jobs/${jobId}/captions?speaker_channel=1`,
```
amikofalvy, line 49: yep this looks exactly like the python test suite. You were right.

kbridbur: Unfortunately :/

README.md
amikofalvy: Let's show another example, one with the param and one without. Add a comment to each to show why we are showing that example.

amikofalvy (approved):
:shipit:

PR #20, opened 2019-08-26
AISDK-55: Fix captiontype export

kbridbur:


src/models/async/CaptionTypes.ts
```diff
@@ -1,4 +1,6 @@
-export const enum CaptionTypes {
+enum CaptionType {
     SRT = 'application/x-subrip',
     VTT = 'text/vtt'
-}
\ No newline at end of file
+}
+
+export default CaptionType;
```
kbridbur: Node cannot export objects normally so we need to do this. From a user perspective it is the same because we change this default export to a named one in the index.ts file

amikofalvy: I think we want 
```
export enum CaptionTypes {}
```

and we can go back to using

```
import { CaptionTypes } from './models/async/CaptionTypes';
```

src/models/async/CaptionTypes.ts
```diff
@@ -1,4 +1,4 @@
-export const enum CaptionTypes {
+export enum CaptionType {
```
amikofalvy, line 1: File is still named with the plural consider making them the same plurality. Also fix in index.ts.

amikofalvy (approved):
Fix plural file name then :shipit:

PR #23, opened 2019-09-15
AISDK-61: Add user agent and metadata

kbridbur:


src/streaming-client.ts
```diff
@@ -53,14 +56,20 @@ export class RevAiStreamingClient extends EventEmitter {
      * Begins a streaming connection. Returns a duplex
      * from which the user can read responses from the api and to which the user
      * should write their audio data
+     * @param metadata (optional) Metadata to be associated with the streaming job
      *
      * @returns BufferedDuplex. Data written to this buffer will be sent to the api
      * Data received from the api can be read from this duplex
      */
-    public start(): Duplex {
+    public start(metadata?: string): Duplex {
```
amikofalvy: For extensibility, we may want to make this parameter an object with a metadata property.

The object could be called SessionConfig.

This way as we add more configuration settings the ordering of parameters won't matter.

amikofalvy (changes requested):
One comment

src/models/streaming/SessionConfig.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Configuration for streaming job
+ */
+export class SessionConfig {
+    metadata?: string;
+
+    /**
+     * @param metadata (Optional) metadata to be associated with the streaming job
```
amikofalvy, line 8: Can the customer skip the constructor and just use an inline object?

Example:

`start({ metadata: "my metadata" })`?

src/streaming-client.ts
```diff
@@ -53,14 +57,20 @@ export class RevAiStreamingClient extends EventEmitter {
...
      * @returns BufferedDuplex. Data written to this buffer will be sent to the api
      * Data received from the api can be read from this duplex
      */
-    public start(): Duplex {
+    public start(config?: SessionConfig): Duplex {
         let url = this.baseUrl +
             `?access_token=${this.accessToken}` +
-            `&content_type=${this.config.getContentTypeString()}`;
+            `&content_type=${this.config.getContentTypeString()}` +
+            `&user_agent=${encodeURIComponent(`RevAi-NodeSDK/${sdkVersion}`)}`;
```
amikofalvy, line 69: Please double check on the server side that the uri is properly decoded.

test/integration/src/client-helper.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     getTranscribedJobId: (jobList) => {
         var completedJobId;
         for(job of jobList) {
-            if(job.status === 'transcribed') {
+            if(job.status === 'transcribed' && job.type != 'stream') {
```
amikofalvy: Do we have status enums defined? Should we?

src/models/streaming/SessionConfig.ts
kbridbur: Yes they should be able to

test/integration/src/client-helper.js
kbridbur: We do not have a status enum defined. I can't remember what our reasoning was for that. I think it's a good idea

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "revai-node-sdk",
-  "version": "2.1.0",
+  "version": "2.1.1",
```
kennygea: should this be version `2.2.0`? I believe the last number is bug fixes, whereas the minor number is for non-breaking features

src/models/streaming/SessionConfig.ts
```diff
@@ -0,0 +1,15 @@
...
+     */
+    constructor(
+        metadata?: string
+    ) {
```
kennygea, line 12: is this how it's supposed to be formatted in JS?

kbridbur: yes, we run a 3rd party linter on the project as part of the travis ci pipeline and this is what it accepts :)

src/streaming-client.ts
kennygea: Should ``RevAi-NodeSDK/${sdkVersion}`` be defined as a constant somewhere?

test/integration/test/captions.test.js
```diff
@@ -1,12 +1,24 @@
...
 const configHelper = require('../src/config-helper');
+const InstanceStateError = require('../../../src/models/RevAiApiError').InvalidStateError;
 
 test('Can get srt captions', async (done) => {
     const client = clientHelper.getClient(configHelper.getApiKey());
     const jobList = await client.getListOfJobs();
     const jobId = clientHelper.getTranscribedJobId(jobList);
     expect(jobId).toBeDefined();
-    const captionsStream = await client.getCaptions(jobId);
+    var captionStream = null;
```
kennygea, line 10: I don't get this change at all, what does it have to do with the issue?

test/unit/streaming-client.spec.ts
```diff
@@ -22,13 +26,46 @@ describe('streaming-client', () => {
...
     describe('start', () => {
         it('Connects to api with parameters', () => {
+            // Arrange
+            const config = new SessionConfig('my metadata');
+
+            // Act
+            const res = sut.start(config);
+
+            // Assert
+            expect(mockClient.connect).toBeCalledWith(`wss://api.rev.ai/speechtotext/v1alpha/stream?` +
```
kennygea: super minor, this url can probably be moved to a constant

src/models/async/JobStatus.ts
```diff
@@ -0,0 +1,5 @@
+export enum JobStatus {
```
kennygea: status is not specific to `async` right? I think it should be directly in the `/models/` namespace

src/models/RevAiApiJob.ts
```diff
@@ -1,6 +1,10 @@
+import { JobStatus } from './async/JobStatus';
+import { JobType } from './JobType';
+
 export interface RevAiApiJob {
     id: string;
-    status: string;
+    status: JobStatus;
```
kennygea, line 6: is this backwards compatible? Like i.e if I compare the enum to a string, does it work? I would add a unit test to check for backwards compatibility

test/integration/test/captions.test.js
kbridbur: Nothing, it's irrelevant to the issue but was causing certain integration tests which should'nt be failing to fail. The integration test picks a job at random (that is transcribed and async) and tries to get captions output. In some cases this job was a speaker channel job so the call would fail

kbridbur: It's a small fix so I just made it. Do you think it needs it's own issue?

src/models/RevAiApiJob.ts
kbridbur: Yes it is backwards compatible

test/unit/api-client.spec.ts
```diff
@@ -1,332 +0,0 @@
-import { RevAiApiClient } from '../../src/api-client';
```
kbridbur, line 1: This was supposed to be removed since this file was split into multiple but it got readded.

test/unit/backwards_compatibility.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { JobStatus } from '../../src/models/JobStatus';
+import { JobType } from '../../src/models/JobType';
+
+describe('test items for backwards compatibility', () => {
```
kbridbur, line 4: Adding this test file for general backwards compatibility testing. Test names in this class will follow the format
`{version change was introduced}+, {test description}`

src/streaming-client.ts
kbridbur: We don't really have a global constants file in the SDK and I don't think it's necessary for now. I don't forsee us having more than 2 uses of this constant string anytime in the near future

test/integration/test/captions.test.js
kennygea: Not really, just make sure to make a comment about off-pr changes so people know what's going on

test/unit/backwards_compatibility.spec.ts
kennygea: I was think more of a single unit test in an existing class, but this works as well.

kennygea (approved):
This looks good from my end, you can ping @amikofalvy to approve as well

test/integration/test/captions.test.js
```diff
@@ -1,12 +1,24 @@
...
+    try {
+        captionsStream = await client.getCaptions(jobId);
+    }
+    catch (error){
```
amikofalvy:
```suggestion
    catch (error) {
```

kbridbur: ~:0 I'm surprised the linter didn't catch this. Will fix~

We don't lint the tests because they're not part of the distribution

amikofalvy (approved):

PR #24, opened 2019-09-26
QAT-460

Chanc3:
* Added beforeAll methods to tests that rely on having `transcribed` jobs available
* Removed the fake timer property from the integration tests to allow the use of setInterval polling

The beforeAll method has the potential to run for about 10 minutes. I noticed that jobs can take anywhere from 3 - 10+ minutes to complete so I opted for the higher value. If this is unreasonable let me know and we can adjust it.

### DIFF
--- a/test/integration/test/captions.test.js
+++ b/test/integration/test/captions.test.js
@@ -1,9 +1,33 @@
 const clientHelper = require('../src/client-helper');
 const configHelper = require('../src/config-helper');
+const JobStatus = require('../../../dist/src/models/JobStatus').JobStatus;
+const client = clientHelper.getClient(configHelper.getApiKey());
 const InstanceStateError = require('../../../src/models/RevAiApiError').InvalidStateError;
 
+beforeAll(async (done) => {
+    const jobList = await client.getListOfJobs();
+    var jobId;
+    if(jobList != undefined) {
+        jobId = clientHelper.getTranscribedJobId(jobList);
+    }
+    if(jobId == undefined) {
+        const job = await client.submitJobUrl('https://www.rev.ai/FTC_Sample_1.mp3');
+        jobId = job.id;
+    }
+    var count = 0;
+    var intervalObject = setInterval(function(){ 
+        count++;
+        (async () => {
+            const jobDetails = await client.getJobDetails(jobId);
+            if (jobDetails.status == JobStatus.Transcribed || count >= 40) { 
+                clearInterval(intervalObject);
+                done();
+            }
+        })()
+    }, 15000);
+}, 600000)
+
 test('Can get srt captions', async (done) => {
-    const client = clientHelper.getClient(configHelper.getApiKey());
     const jobList = await client.getListOfJobs();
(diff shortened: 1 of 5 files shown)


test/integration/test/captions.test.js
```diff
@@ -1,9 +1,33 @@
 const clientHelper = require('../src/client-helper');
 const configHelper = require('../src/config-helper');
+const JobStatus = require('../../../dist/src/models/JobStatus').JobStatus;
+const client = clientHelper.getClient(configHelper.getApiKey());
 const InstanceStateError = require('../../../src/models/RevAiApiError').InvalidStateError;
 
+beforeAll(async (done) => {
+    const jobList = await client.getListOfJobs();
+    var jobId;
+    if(jobList != undefined) {
```
kbridbur: should be `!==` in javascript we generally want to use `===` and `!==` over `!=` and `==` because the triple ones also enforce type

test/integration/test/captions.test.js
```diff
@@ -1,9 +1,33 @@
...
+        jobId = clientHelper.getTranscribedJobId(jobList);
+    }
+    if(jobId == undefined) {
```
kbridbur: `===`

test/integration/test/captions.test.js
```diff
@@ -1,9 +1,33 @@
...
+        const job = await client.submitJobUrl('https://www.rev.ai/FTC_Sample_1.mp3');
+        jobId = job.id;
+    }
+    var count = 0;
+    var intervalObject = setInterval(function(){ 
+        count++;
+        (async () => {
+            const jobDetails = await client.getJobDetails(jobId);
+            if (jobDetails.status == JobStatus.Transcribed || count >= 40) {
```
kbridbur: I'm confused about the purpose of the count. Won't beforeAll timeout before we get to it?

test/integration/test/list.test.js
```diff
@@ -1,17 +1,26 @@
 const clientHelper = require('../src/client-helper');
 const configHelper = require('../src/config-helper');
 const RevAiApiJob = require('../../../dist/src/models/RevAiApiJob')
+const client = clientHelper.getClient(configHelper.getApiKey());
+
+beforeAll(async (done) => {
+    const jobList = await client.getListOfJobs();
+    if(jobList == undefined || jobList.length < 2) {
```
kbridbur: `===`

test/integration/test/list.test.js
```diff
@@ -1,17 +1,26 @@
...
+        for(var i = 0; i < 2; i++) {
```
kbridbur: minor: you could just call `client.submitJobUrl` twice instead of looping. I think it's preference though

kbridbur (changes requested):
Looks mostly good, a couple questions and minor comments

test/integration/test/captions.test.js
Chanc3: Yea you are right. I'll remove the count.

PR #26, opened 2019-09-30
QAT-457

Chanc3:
Added a streaming integration test to run outside of jest


test/integration/test/stream.js
```diff
@@ -0,0 +1,73 @@
...
+const assert = require('assert');
+
+(async (done) => {
+    const audioConfig = new AudioConfig("audio/x-raw", 'interleaved', 16000, 'S16LE', 1);
+    const client = new RevAiStreamingClient(configHelper.getApiKey(), audioConfig);
+    client.baseUrl = `wss://${configHelper.getBaseUrl()}/speechtotext/v1alpha/stream`;
+        
+    client.on('close', (code, reason) => {
+        console.log(`Connection closed, ${code}: ${reason}`);
+        assert.equal(code, 1000, `Expected close code to be [1000] but was ${code}`);
```
kbridbur: Can we move these to another helper like the other asserts? Would be nice to have all the assertions at the bottom of the file in an easy to find place

test/integration/test/stream.js
```diff
@@ -0,0 +1,73 @@
...
+        console.log(`Connection failed with error: ${error}`);
+    });
+    client.on('connect', connectionMessage => {
+        console.log(`Connected with job id: ${connectionMessage.id}`);
+    });
+
+    var stream = client.start();
+
+    stream.on('data', data => {
+        if(data.type === 'partial') {
```
kbridbur:
```suggestion
        if (data.type === 'partial') {
```

test/integration/test/stream.js
```diff
@@ -0,0 +1,73 @@
...
+    var file = fs.createReadStream('../resources/english_test.raw');
+
+    file.on('end', () => {
+        client.end();
+    });
+
+    file.once('readable', () => {
+        file.pipe(stream);
+    });
+
```
kbridbur: Extra newline

test/integration/test/stream.js
```diff
@@ -0,0 +1,73 @@
...
+function assertFinalHypothesis(final) {
+    assert.ok(final.ts < final.end_ts);
+    final.elements.forEach(element => {
+        if(element.type === 'punct') {
+            assert.notEqual(element.value, 'undefined', 'Expected element value to not be [undefined]: ' + JSON.stringify(element));
+        } else {
+            assert.equal(element.type, 'text', 'Expected element type to be [text]: ' + JSON.stringify(element));
+            assert.notEqual(element.value, 'undefined', 'Expected element value to not be [undefined]: ' + JSON.stringify(element));
+            assert.ok(element.ts < element.end_ts, 'Expected starting timestamp value is not less than ending timestamp: ' + JSON.stringify(element));
+            assert.equal(element.confidence, 1, 'Expected confidence to be [1]:' + JSON.stringify(element));
```
kbridbur: I don't know if this is always true

kbridbur (commented):
Looks good, a few small comments

test/integration/test/stream.js
```diff
@@ -0,0 +1,82 @@
...
+        console.log(`Connection closed, ${code}: ${reason}`);
+        assertCloseCodeAndReason(code, reason);
+        printPassStatement();
+        return;
+    });
+    client.on('httpResponse', code => {
+        console.log(`Streaming client received http response with code: ${code}`);
+    });
+    client.on('connectFailed', error => {
+        console.log(`Connection failed with error: ${error}`);
```
kbridbur: We should error and fail in this case

test/integration/test/stream.js
```diff
@@ -0,0 +1,82 @@
...
+    client.baseUrl = `wss://${configHelper.getBaseUrl()}/speechtotext/v1alpha/stream`;
+        
+    client.on('close', (code, reason) => {
+        console.log(`Connection closed, ${code}: ${reason}`);
+        assertCloseCodeAndReason(code, reason);
+        printPassStatement();
+        return;
+    });
+    client.on('httpResponse', code => {
+        console.log(`Streaming client received http response with code: ${code}`);
```
kbridbur: Error and fail here

test/integration/test/stream.js
```diff
@@ -0,0 +1,82 @@
...
+    stream.on('data', data => {
+        if (data.type === 'partial') {
+            assertPartialHypothesis(data);
+        } else if (data.type === 'final') {
+            assertFinalHypothesis(data);
+        }
+    });
+
+    stream.on('error', (error) => {
+        console.log(`Streaming error occurred: ${error}`);
```
kbridbur: Here as well

test/integration/test/stream.js
```diff
@@ -0,0 +1,82 @@
...
+    });
+    client.on('connect', connectionMessage => {
+        console.log(`Connected with job id: ${connectionMessage.id}`);
+    });
+
+    var stream = client.start();
+
+    stream.on('data', data => {
+        if (data.type === 'partial') {
+            assertPartialHypothesis(data);
```
kbridbur, line 31: Will this test still pass if we don't get any partials or finals? I feel like it will

Chanc3: Good catch. It will fail but later down the road during the close assertions but that would not be an accurate failure. I can add an else conditional here that explicitly fails the test in the case neither are present.

package.json
```diff
@@ -41,6 +41,7 @@
     "@types/form-data": "^2.2.1",
     "@types/jest": "^23.3.13",
     "@types/node": "^10.12.18",
+    "colors": "^1.4.0",
```
kbridbur: I think we should remove this package. The colors are cool but don't add anything for users

test/integration/test/stream.js
```diff
@@ -0,0 +1,83 @@
+const {AudioConfig} = require('../../../dist/src/models/streaming/AudioConfig');
+const {RevAiStreamingClient} = require('../../../dist/src/streaming-client');
+const configHelper = require('../src/config-helper');
+const fs = require('fs');
+const assert = require('assert');
+const colors = require('colors');
+
+(async (done) => {
```
kbridbur: I dont think this `done` should be there, probably a relic from when we wanted to use jest

kbridbur (approved):
:shipit:

test/integration/test/stream.js
```diff
@@ -0,0 +1,83 @@
+const {AudioConfig} = require('../../../dist/src/models/streaming/AudioConfig');
+const {RevAiStreamingClient} = require('../../../dist/src/streaming-client');
+const configHelper = require('../src/config-helper');
+const fs = require('fs');
+const assert = require('assert');
+const colors = require('colors');
```
kbridbur: Get rid of this require

PR #28, opened 2019-10-01
AISDK-62: Add submit method for buffer and readable stream

kbridbur:


src/api-client.ts
```diff
@@ -57,9 +57,25 @@ export class RevAiApiClient {
...
 
-    async submitJobLocalFile(filename: string, options?: RevAiJobOptions): Promise<RevAiApiJob> {
+    async submitJobFileStream(
+        fileStream: Buffer | Readable,
+        filename: string,
+        options?: RevAiJobOptions
+    ): Promise<RevAiApiJob> {
         let payload = new FormData();
-        payload.append('media', fs.createReadStream(filename));
+        console.log(fileStream);
```
kbridbur: Will remove

test/integration/test/captions.test.js
```diff
@@ -12,7 +12,8 @@ test('Can get srt captions', async (done) => {
         captionsStream = await client.getCaptions(jobId);
     }
     catch (error) {
-        if (error instanceof InvalidStateError){
+        console.log(error instanceof InvalidParameterError);
```
kbridbur: will remove

test/integration/test/job.test.js
```diff
@@ -14,7 +15,29 @@ test('Can submit url', async () => {
...
+    options.metadata = 'Node sdk submit buffer';
+    const fileStream = fs.readFile('./test/integration/resources/test_mp3.mp3', async (err, data) => {
+        const job = await client.submitJobFileStream(data, 'test_mp3.mp3', options);
+        expect(job.status).toBe('in_progress');
+        expect(job.id).not.toBeNull();
+        done();
+    });
+});
+
+test('Can submit ReadableStream', async (done) => {
```
kbridbur: this test is currently timing out, investigating why

src/api-client.ts
```diff
@@ -57,9 +57,25 @@ export class RevAiApiClient {
             { 'Content-Type': 'application/json' }, 'json', options);
     }
 
-    async submitJobLocalFile(filename: string, options?: RevAiJobOptions): Promise<RevAiApiJob> {
+    async submitJobFileStream(
+        fileStream: Buffer | Readable,
```
kbridbur: `Buffer` isn't exactly a stream but it can be treated in the same way. Not sure if I should change the method name somehow

dkokotov: perhaps `submitJobWithAudioData(audioData: Buffer | Readable)`?

using `FileStream` for method name / property name is somewhat misleading because it doesn't have to come rom a file (presumably the point of the method is to allow you to not have to get the data from a file)

src/api-client.ts
```diff
@@ -57,9 +57,25 @@ export class RevAiApiClient {
...
+        payload.append('media', fileStream, { filename: filename });
+        if (options) {
+            payload.append('options', JSON.stringify(options));
+        }
+
+        return await this.apiHandler.makeApiRequest<RevAiApiJob>('post', `/jobs`,
+            payload.getHeaders(), 'json', payload);
+    }
+
+    async submitJobLocalFile(filepath: string, options?: RevAiJobOptions): Promise<RevAiApiJob> {
```
dkokotov, line 134: Unrelated to the PR, but i noticed there is no doc comments for any of the methods, is that deliberate?

in the python SDK we have doc comments - it would seem nice to have them for node as well

kbridbur: Agreed, I think it's just been an oversight. The streaming-client was more recently written and does have doc comments. I'll file an issue

src/api-client.ts
```diff
@@ -57,9 +99,41 @@ export class RevAiApiClient {
             { 'Content-Type': 'application/json' }, 'json', options);
     }
 
-    async submitJobLocalFile(filename: string, options?: RevAiJobOptions): Promise<RevAiApiJob> {
+    /**
+     * See https://www.rev.ai/docs#operation/SubmitTranscriptionJob
+     * Submit audio data for transcription.
+     * @param fileStream Audio data to be transcribed.
+     * @param filename Name of file associated with audio, file extension is used to determine type of audio.
```
kennygea: maybe make this optional? I would think from `buffer`, there isn't always a `filename` present.

Uncertain what's the best experience right now, I just think it's kind of confusing to have to pass a filename when the name of the method seems to indicate all we need is a buffer of data

src/api-client.ts
```diff
@@ -14,20 +14,48 @@ const enum TranscriptContentTypes {
...
+     * @param id Id of job whose details are to be retrieved
+     * @returns Job details
+     */
     async getJobDetails(id: string): Promise<RevAiApiJob> {
         return await this.apiHandler.makeApiRequest<RevAiApiJob>( 'get', `/jobs/${id}`, {}, 'json');
     }
 
+    /**
+     * See https://www.rev.ai/docs#operation/GetListOfJobs
+     * Get a list of transcription jobs submitted within the last week in reverse chronological order
```
kennygea, line 52: maybe put `(Last submitted first)`, just to clarify what reverse chronological order means

README.md
```diff
@@ -46,6 +46,9 @@ var job = await client.submitJobLocalFile("./path/to/file.mp4");
 
 // or submit via a public url
 var job = await client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3");
+
+// or from audio data, this requires that you also provide a filename
+var job = await client.submitJobAudioData("Insert Buffer or ReadableStream", "filename.mp3");
```
kennygea: What do you think of having the example open a local file and pass the buffer through here instead?

src/api-client.ts
kbridbur: filename needs to be provided otherwise the API won't accept the "media file" since it has no name or (more importantly) content type

kennygea: If the filename is not provided by the user, we can generate it ourselves

kennygea: I guess the content-type can be tricky

kbridbur: I originally thought content type was coming from filename. It is not, we can make this optional and just use a generic, extensionless filename

README.md
```diff
@@ -46,6 +46,10 @@ var job = await client.submitJobLocalFile("./path/to/file.mp4");
 
 // or submit via a public url
 var job = await client.submitJobUrl("https://www.rev.ai/FTC_Sample_1.mp3");
+
+// or from audio data, this requires that you also provide a filename
```
kennygea: this comment is not true anymore right?

kennygea (approved):
One more comment, otherwise looks good to me :shipit:

PR #29, opened 2019-10-07
AISDK-65: Remove content length limit from axios

kbridbur:


src/api-request-handler.ts
```diff
@@ -25,6 +25,7 @@ export class ApiRequestHandler {
     constructor (url: string, accessToken: string) {
         this.instance = axios.create({
             baseURL: url,
+            maxContentLength: Infinity,
```
kbridbur, line 28: Default was 10MB. I haven't tested a massive file but after this change I was able to submit a 20MB file with no issue

amikofalvy: We limit on the server side anyway. so setting this to infinity shouldn't be an issue.

amikofalvy (approved):

PR #32, opened 2019-10-28
AISDK-68: Node 12 support

kbridbur:


.travis.yml
```diff
@@ -2,43 +2,64 @@ language: node_js
...
 jobs:
   include:
-    - stage: build
+    - &building
+      stage: build
       script: npm run build
       node_js: "8"
-    - stage: build
-      script: npm run build
+    - <<: *building
```
kbridbur: This syntax isn't documented anywhere that i could find -_- , but you can use this page to see what it compiles down to 
https://config.travis-ci.com/explore

kbridbur: Essentially copies the stage at the `&` but you can override variables

amikofalvy: I think this is yaml syntax.

kbridbur: Makes sense. Good to know

amikofalvy: We use similar syntax in our definition of the CloudFormation parameter YAML files for REVAI and RSS.

.travis.yml
```diff
@@ -2,43 +2,64 @@ language: node_js
...
         - secure: "rMIaZbjPW6vevo5f550j+N3fN313oOo94SCL15zzwqiNWjctIPCXW7wr1XST/oEU/+YM+UvhGZ12aXQf/Bh+ptxbvqQ7N6II64qbYRGqKBQqt6lbaUYQvoKJHbxXV8wBMxXkfzrpwgHiVZ3Yp6SCsa7OaHTasa+H19svsIdyA6WgHWDGpAxe0gFnMlAX0bHk/T8Kbke3gP9k+V0rEzIrqihgHZtldJfd4dcAZGB/XC6+AyNrYHKdMjFTft3YZODcb5xXvBsaUhFCDylqNA997GtKHQbzcF5imFGxTHRzBXDceAKL6cs3pF604K5oy3WgDiU5JJMihO0jbSLbpv2VLj+CDAPSL97pKJmOFn0R/HKOn1uxjO5r31q+73EZwZu9uE6dCgj06tv5NLQ9TrcP3G0EA2l7yaCnm/tPwnYXdLab0qtCShnAFJIjaf0tQcqhM9BpAWNVuRIQjYsHOADn3GHUgRiWWg1MYl6EN8ORemenkvWTzIq6/5LZ+ycrjFqb1+paRBRauud97kMoR6AwHJnjayviCgEvY+51sDDZl9NBR6bH4WcJfQcWbvj9JWWBOnjrjGdeOsj2wTCY1IDAAtL/jXDJ73vN9hhJtZ35xegBY66D7g53ApTkdFLhWdQ7E0ngkoCHbTmU8Z5WeGSInnsdTOz7GbjmgOWqG6T5vOM="
-    - stage: Stage Integration
+    - <<: *test-integration
+      node_js: "10"
+    - <<: *test-integration
+      node_js: "12"
+    - &stage-integration
+      stage: stage integration
       script:
```
amikofalvy: I think you can get rid of the script property because it is already defined in `&test-integration` and it is the same for all other properties.

amikofalvy (approved):
One suggestion to clean up the yaml, otherwise looks good.

.travis.yml
kbridbur: If I get rid of the script property for this stage it will inherit the script from the top of the file (so just run unit tests). using the << syntax I think I can make all of these inherit each other. I'll try that

.travis.yml
```diff
@@ -2,43 +2,59 @@ language: node_js
...
+  - "12"
 
 script: npm run unit-test
 
+common: &integration-common
+  script:
+    - npm run build
+    - node build_release_package_json.js
+    - npm run integration-test
+  node_js: "8"
```
amikofalvy: I think you can remove the definition of node_js in common. `node_js` should be defined for every step independently.

.travis.yml
```diff
@@ -2,43 +2,59 @@ language: node_js
...
+
 jobs:
   include:
-    - stage: build
+    - stage: lint
+      script: npm run lint
+    - &building
+      stage: build
       script: npm run build
```
amikofalvy: you can extract the stage and script property into the `common:` definition
```
common:
 buildstage: &build
  stage: build
  script: npm run build

...

-stage <<: *build
 node_js: 8
-stage <<: *build
 node_js: 10
-stage <<: *build
 node_js: 12

```

amikofalvy: The benefit is no default values for node_js so we can be explicit that the difference between these stage definition is the node versions

.travis.yml
```diff
@@ -2,43 +2,59 @@ language: node_js
 node_js: 
   - "8"
   - "10"
+  - "12"
 
 script: npm run unit-test
 
+common: &integration-common
```
amikofalvy: Instead of common, I would call this `Defaults`.

I would put all other shared properties as subproperties of `Defaults`. This way there is only one property that travis can confuse for one of its special keywords.

.travis.yml
```diff
@@ -2,43 +2,59 @@ language: node_js
...
         - secure: "ExErbWSJNpnCSh8NGczBfCY+8O0kGrCyXFf16lufioVWQVUT3uUNQ7Th/FCq84ReYSWGHnEeLKGMTPq8B+wfvWADOb1Lnu2Sz3ADXPpeGurdRmveBYrxmh7jrXeCCb8Qil2E+kNn9MDnxDUoCXDxw4c/hLLiV9MpTj9+22gc846ti4E/tWJrZaGfJnO3zPfJmar19mwQRi8sx98Hm3M/565rdDGDCOgrZSaZYsdF+/MCZt9mM9Da1OZi2IGDdCbgrJ0ERoP+v/b3KqLP9H+7G1CafPoRPAgHur2OBcXXk+dtsid+yymcc/ZQKsIpWt9TnVj0BlKM0cVG79q7h2RG4qQsPkgaSX5JzZngatm7QKeWkQtqXNWvVuihNkA3XbxtJPUaBvsHH/GleKutQ7I8elBhY4BYHM7l4aALUgONFibE3EeF9VU/oGHDxGoD/je2GTxBk4XiuKpn9Wtj7Fy9sZ+fvqL1YSSWjWEHvR1vKYF3GPu5VLMs5a1y/DkTvipY86RFAh6HOAuvOD0sLGYdc/f50okJ5HJiUoC5w/m2K0z7/mw8s9sPuU/UmzUlOfQjtdpupDwgGpOAPRL7uyyEWrgDZZimLTeNdf7qIYdycrh9ODNRAeishdT4ewhLLsmQrbOqIFloL+oJ1Ca0EueqLRLbkVHc39Frw4apyman8EA="
+    - <<: *test-integration
```
amikofalvy, line 52: This is a nice way not to duplicate property definitions :)

amikofalvy (approved):

PR #34, opened 2019-11-14
AISDK-67: Add custom vocab parameter to the streaming client Node

KostasRev:


src/models/streaming/SessionConfig.ts
```diff
@@ -3,13 +3,18 @@
  */
 export class SessionConfig {
```
amikofalvy, line 4: For the Github PR title, you don't need the feature prefix.

kbridbur (approved):
Add a test where customVocabId is passed in the session config. Other than that looks good
:shipit:

PR #41, opened 2020-03-04
AISDK-92: Add filter_profanity to the Node SDK

KostasRev:
Tested manually locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,3 +188,17 @@ var customVocabularySubmission = await client.submitCustomVocabularies(customVoc
 // Get information regarding the Custom Vocabulary submission and its progress
 var customVocabularyInformation = await client.getCustomVocabularyInformation(customVocabularySubmission.id)
 ```
+
+# For Rev.ai NodeSDK Developers
+
+After cloning and installing required npm modules, you should follow these practices when developing:
+
+1. Use the scripts defined in [package.json](https://github.com/revdotcom/revai-node-sdk/tree/develop/package.json) in this manner `npm run [command_name]`:
+    1. `lint` checks that you are not violating any code style standards. This ensures our code's style quality stays high improving readability and reducing room for errors.
+    2. `build` transpiles the typescript into javascript with the options specified in [tsconfig.json](https://github.com/revdotcom/revai-node-sdk/tree/develop/tsconfig.json)
+    3. `unit-test` runs our unit tests which live in the [unit test directory](https://github.com/revdotcom/revai-node-sdk/tree/develop/test/unit).
+    * Note that `integration-test` is currently configured to work with a certain account specified in our continuous integration build environment, as such for now you can check the automated continuous integration checks to pass the integration tests. This will change in the future.
+2. Add any relevant test logic if you add or modify any features in the source code and check that the tests pass using the scripts mentioned above.
+3. Update the examples provided to illustrate any relevant changes you made, and check that they work properly.
+    * Note that you will have to copy the output of the `build` script into the `examples/node_modules/revai-node-sdk` directory in order to use your changes in the examples. On Unix, this can be simply done with the following command when in the root directory: `$ cp -r dist/src examples/node_modules/revai-node-sdk/`.
+4. Update the documentation to reflect any relevant changes and improve the development section.
--- a/README.md
+++ b/README.md
@@ -199,6 +199,6 @@ After cloning and installing required npm modules, you should follow these pract
     3. `unit-test` runs our unit tests which live in the [unit test directory](https://github.com/revdotcom/revai-node-sdk/tree/develop/test/unit).
     * Note that `integration-test` is currently configured to work with a certain account specified in our continuous integration build environment, as such for now you can check the automated continuous integration checks to pass the integration tests. This will change in the future.
 2. Add any relevant test logic if you add or modify any features in the source code and check that the tests pass using the scripts mentioned above.
-3. Update the examples provided to illustrate any relevant changes you made, and check that they work properly.
-    * Note that you will have to copy the output of the `build` script into the `examples/node_modules/revai-node-sdk` directory in order to use your changes in the examples. On Unix, this can be simply done with the following command when in the root directory: `$ cp -r dist/src examples/node_modules/revai-node-sdk/`.
+3. Update the examples provided to illustrate any relevant changes you made, and check that they work properly with your changed local `revai-node-sdk`.
+    * One way to use your changed local package in the examples is to copy the output of the `build` script into the `examples/node_modules/revai-node-sdk`. On Unix, this can be simply done with the following command when in the root directory: `$ cp -r dist/src examples/node_modules/revai-node-sdk/`.
 4. Update the documentation to reflect any relevant changes and improve the development section.
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ After cloning and installing required npm modules, you should follow these pract
     1. `lint` checks that you are not violating any code style standards. This ensures our code's style quality stays high improving readability and reducing room for errors.
     2. `build` transpiles the typescript into javascript with the options specified in [tsconfig.json](https://github.com/revdotcom/revai-node-sdk/tree/develop/tsconfig.json)
     3. `unit-test` runs our unit tests which live in the [unit test directory](https://github.com/revdotcom/revai-node-sdk/tree/develop/test/unit).
-    * Note that `integration-test` is currently configured to work with a certain account specified in our continuous integration build environment, as such for now you can check the automated continuous integration checks to pass the integration tests. This will change in the future.
+    * Note that `integration-test` is currently configured to work with a certain account specified in our continuous integration build environment, as such for now you can check the automated continuous integration checks to pass the integration tests.
 2. Add any relevant test logic if you add or modify any features in the source code and check that the tests pass using the scripts mentioned above.
 3. Update the examples provided to illustrate any relevant changes you made, and check that they work properly with your changed local `revai-node-sdk`.
     * One way to use your changed local package in the examples is to copy the output of the `build` script into the `examples/node_modules/revai-node-sdk`. On Unix, this can be simply done with the following command when in the root directory: `$ cp -r dist/src examples/node_modules/revai-node-sdk/`.
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ var job = await client.submitJobAudioData(stream, "file.mp3");
 `job` will contain all the information normally found in a successful response from our
 [Submit Job](https://www.rev.ai/docs#operation/SubmitTranscriptionJob) endpoint.
 
-If you want to get fancy, both send job methods can take a `RevAiJobOptions` object which contains fields for `metadata`, `callback_url`, `skip_diarization`,`skip_punctuation`, `speaker_channels_count` and `custom_vocabularies` as optional parameters. These are also described in the request body of the [Submit Job](https://www.rev.ai/docs#operation/SubmitTranscriptionJob) endpoint.
+If you want to get fancy, both send job methods can take a `RevAiJobOptions` object which contains fields for `metadata`, `callback_url`, `skip_diarization`,`skip_punctuation`, `speaker_channels_count`, `filter_profanity` and `custom_vocabularies` as optional parameters. These are also described in the request body of the [Submit Job](https://www.rev.ai/docs#operation/SubmitTranscriptionJob) endpoint.
 
 ### Checking your job's status
 
--- a/examples/async_transcribe_local_media_file.js
+++ b/examples/async_transcribe_local_media_file.js
@@ -15,6 +15,8 @@ const token = require('./config/config.json').access_token;
         metadata: "InternalOrderNumber=123456789",
         callback_url: "https://jsonplaceholder.typicode.com/posts",
         skip_diarization: false,
+        skip_punctuation: false,
+        speaker_channels_count: 1,
         custom_vocabularies: [{
             phrases: [
                 "add",
@@ -22,7 +24,8 @@ const token = require('./config/config.json').access_token;
                 "vocabularies",
                 "here"
             ]
-        }]
+        }],
+        filter_profanity: true
     };
 
     // Media may be submitted from a local file
--- a/examples/async_transcribe_local_media_file.js
+++ b/examples/async_transcribe_local_media_file.js
@@ -16,7 +16,7 @@ const token = require('./config/config.json').access_token;
         callback_url: "https://jsonplaceholder.typicode.com/posts",
         skip_diarization: false,
         skip_punctuation: false,
-        speaker_channels_count: 1,
+        speaker_channels_count: null,
         custom_vocabularies: [{
             phrases: [
                 "add",
--- a/examples/async_transcribe_local_media_file.js
+++ b/examples/async_transcribe_local_media_file.js
@@ -25,7 +25,7 @@ const token = require('./config/config.json').access_token;
                 "here"
             ]
         }],
-        filter_profanity: true
+        filter_profanity: false
     };
 
     // Media may be submitted from a local file
--- a/examples/async_transcribe_media_from_url.js
+++ b/examples/async_transcribe_media_from_url.js
@@ -15,8 +15,6 @@ const token = require('./config/config.json').access_token;
         metadata: "InternalOrderNumber=123456789",
         callback_url: "https://jsonplaceholder.typicode.com/posts",
         skip_diarization: false,
-        skip_punctuation: false,
-        speaker_channels_count: null,
         custom_vocabularies: [{
             phrases: [
                 "add",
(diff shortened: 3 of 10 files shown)


src/models/streaming/SessionConfig.ts
```diff
@@ -12,9 +13,11 @@ export class SessionConfig {
      */
     constructor(
         metadata?: string,
-        customVocabularyID?: string
+        customVocabularyID?: string,
+        filterProfanity?: boolean
```
amikofalvy, line 17: It's important to note in this PR that Kosta and I discussed making SessionConfig an interface to be more inline with the async api JobOptions interface object, but making this change would force a major version change of the sdk so we chose to leave SessionConfig as a class.


We may want to consider separate versioning for the streaming and async clients. It would be a shame to make a major version bump for both if only the streaming client is the one with the breaking change.

examples/streaming_transcribe_from_local_file.js
```diff
@@ -27,8 +27,15 @@ client.on('connect', connectionMessage => {
     console.log(`Connected with job id: ${connectionMessage.id}`);
 })
 
+// Optional config to be provided.
+const sessionConfig = new revai.SessionConfig(
+    /* (optional) metadata */ "this is an example", 
+    /* (optional) custom_vocabulary_id */ null, 
+    /* (optional) filter_profanity */ true
+);
```
k-weng: Very minor - would it better to switch the places of the argument and the comment? Like
```suggestion
// Optional config to be provided.
const sessionConfig = new revai.SessionConfig(
    "this is an example",  /* (optional) metadata */ 
    null,                  /* (optional) custom_vocabulary_id */ 
    true                   /* (optional) filter_profanity */
);
```

test/unit/streaming-client.spec.ts
```diff
@@ -38,12 +38,14 @@ describe('streaming-client', () => {
...
-                `&metadata=${encodeURIComponent(config.metadata)}`
+                `&metadata=${encodeURIComponent(config.metadata)}` +
+                `&custom_vocabulary_id=${encodeURIComponent(config.customVocabularyID)}` +
+                `&filter_profanity=${encodeURIComponent(config.filterProfanity)}`
             );
             expect(mockClient.connect).toBeCalledTimes(1);
         });
 
-        it ('does not add metadata if no config provided', () => {
+        it ('does not add optional parameters if no config provided', () => {
```
amikofalvy, line 48: I think we also need a test for if each optional parameter is null/ not null

examples/async_transcribe_local_media_file.js
```diff
@@ -15,14 +15,17 @@ const token = require('./config/config.json').access_token;
         metadata: "InternalOrderNumber=123456789",
         callback_url: "https://jsonplaceholder.typicode.com/posts",
         skip_diarization: false,
+        skip_punctuation: false,
+        speaker_channels_count: 1,
```
amikofalvy: 1 is not the default value for speaker channels, we should not set this value in the example as it could change how stereo audios are transcribed.

examples/async_transcribe_local_media_file.js
```diff
@@ -15,14 +15,17 @@ const token = require('./config/config.json').access_token;
...
         custom_vocabularies: [{
             phrases: [
                 "add",
                 "custom",
                 "vocabularies",
                 "here"
             ]
-        }]
+        }],
+        filter_profanity: true
```
amikofalvy: Also, this should be set to the assumed default : false

examples/streaming_transcribe_from_local_file.js
```diff
@@ -27,8 +27,15 @@ client.on('connect', connectionMessage => {
     console.log(`Connected with job id: ${connectionMessage.id}`);
 })
 
+// Optional config to be provided.
+const sessionConfig = new revai.SessionConfig(
+    /* (optional) metadata */ "this is an example",
```
amikofalvy: I have not seen the prefixed comment documentation before. is this standard? I would default to putting the comments on the end of the lines.

examples/streaming_transcribe_from_local_file.js
```diff
@@ -27,8 +27,15 @@ client.on('connect', connectionMessage => {
     console.log(`Connected with job id: ${connectionMessage.id}`);
 })
 
+// Optional config to be provided.
+const sessionConfig = new revai.SessionConfig(
+    /* (optional) metadata */ "this is an example", 
+    /* (optional) custom_vocabulary_id */ null, 
+    /* (optional) filter_profanity */ true
```
amikofalvy: And I would use the default value of false here as well.

examples/streaming_transcribe_from_local_file.js
```diff
@@ -27,8 +27,15 @@ client.on('connect', connectionMessage => {
...
+);
+
 // Begin streaming session
-var stream = client.start();
+var stream = client.start(/* sessionConfig */);
```
amikofalvy:
```suggestion
var stream = client.start(sessionConfig);
```

amikofalvy: Use the session config object you created with default values (the only non default value is metadata, but this option does not effect transcription so I believe it to be an acceptable value.

examples/streaming_transcribe_from_local_file.js
amikofalvy:
```suggestion
    /* (optional) metadata */ "my example metadata string", 
```

amikofalvy (changes requested):
I left comments to address. Please review and repost for another round.

examples/streaming_transcribe_from_local_file.js
KostasRev: I'll check to see if it looks better in the IDEs.

test/unit/streaming-client.spec.ts
KostasRev: The not null case is covered by the test in which all configs are provided. I can add tests for the null case however.

examples/async_transcribe_local_media_file.js
KostasRev: I see, I just wanted to display all the potential args. What do you suggest I set it to? Or should I comment it out?

examples/streaming_transcribe_from_local_file.js
KostasRev: Given that both you and @k-weng expressed this sentiment, I'll change this.

test/unit/streaming-client.spec.ts
KostasRev: Addressed.

test/unit/streaming-client.spec.ts
```diff
@@ -38,12 +38,14 @@ describe('streaming-client', () => {
                 `?access_token=${token}` +
                 `&content_type=${audioConfig.getContentTypeString()}` +
                 `&user_agent=${encodeURIComponent(`RevAi-NodeSDK/${sdkVersion}`)}` +
-                `&metadata=${encodeURIComponent(config.metadata)}`
+                `&metadata=${encodeURIComponent(config.metadata)}` +
+                `&custom_vocabulary_id=${encodeURIComponent(config.customVocabularyID)}` +
+                `&filter_profanity=${encodeURIComponent(config.filterProfanity)}`
```
jennywong2129, line 43: In streaming-client.ts line 78 you had a .toString() after the filterProfanity property, I think you should keep this line consistent with that line (either both with .toString() or both without)

test/unit/api-client/api-client.submit.spec.ts
```diff
@@ -36,13 +36,38 @@ describe('api-client job submission', () => {
...
+                skip_punctuation: null,
+                skip_diarization: null,
+                speaker_channels_count: null,
+                filter_profanity: null
+            };
+
+            const job = await sut.submitJobUrl(mediaUrl, null);
+
+            expect(mockHandler.makeApiRequest).toBeCalledWith('post', '/jobs',
+                { 'Content-Type': 'application/json' }, 'json', { media_url: mediaUrl });
```
jennywong2129, line 67: Not sure if we structure java code the same as C#, but maybe have the ); on a new line & each parameter for .toBeCalledWith on a new line as well

KostasRev: It passes the lint check script and seems to be present in other places as well so I think it's fine?

examples/async_transcribe_local_media_file.js
KostasRev: We discussed this and the default value should be `null`. There will need to be an update to the documentation in the near future to clarify this as well.

test/unit/streaming-client.spec.ts
KostasRev: Addressed this in the streaming client!

test/unit/api-client/api-client.submit.spec.ts
```diff
@@ -36,13 +36,38 @@ describe('api-client job submission', () => {
             expect(job).toEqual(jobDetails);
         });
 
+        it('submit job with media url with null options', async () => {
+            const mockHandler = ApiRequestHandler.mock.instances[0];
+            mockHandler.makeApiRequest.mockResolvedValue(jobDetails);
+            const options = {
```
amikofalvy, line 54: options isn't used in this test.

amikofalvy: delete it.

amikofalvy: Do you have tslint enabled on vscode? You should enable it and make sure that you fx any errors.

Also, we may want to add https://palantir.github.io/tslint/rules/no-unused-variable/ option to the tslint.json file in this repo.

test/unit/api-client/api-client.submit.spec.ts
```diff
@@ -112,6 +137,34 @@ describe('api-client job submission', () => {
...
+
+            const job = await sut.submitJobAudioData(fakeStream, null, options);
+
+            const expectedPayload = expect.objectContaining({
+                '_boundary': expect.anything(),
+                '_streams': expect.arrayContaining([expect.anything(),
+                    expect.stringContaining('Content-Disposition: form-data; name="media"')])
+            });
+            const expectedHeader = { 'content-type': expect.stringMatching(/multipart\/form-data; boundary=.+/) };
+            expect(mockHandler.makeApiRequest).toBeCalledWith('post', '/jobs',
```
amikofalvy, line 191: There are no new test assertions in this test.

Either add a relevant assertion or remove this  test.

Additionally, the test case name `submit job with null options` isn't accurate. options isn't null, but the properties  of options are null.

amikofalvy: You should assert that these properties are not present: https://github.com/revdotcom/revai-node-sdk/pull/41/files#diff-55616ed1583fc5edff382eb7d3ddcc36R189-R197

test/unit/api-client/api-client.submit.spec.ts
```diff
@@ -168,6 +224,33 @@ describe('api-client job submission', () => {
...
+                skip_diarization: null,
+                speaker_channels_count: null,
+                filter_profanity: null
+            };
+
+            const job = await sut.submitJobLocalFile(filename, options);
+
+            const expectedPayload = expect.objectContaining({
+                '_boundary': expect.anything(),
+                '_streams': expect.arrayContaining([expect.stringContaining('Content-Type: audio/mpeg'),
```
amikofalvy: Needs the same missing or null property assertion as described above.

test/unit/streaming-client.spec.ts
amikofalvy: Why wouldn't this work in the streaming client, but it works in this test file?

amikofalvy (changes requested):
One more round of review after you fix the tests

test/unit/api-client/api-client.submit.spec.ts
KostasRev: Apparently:
`no-unused-variable is deprecated. Since TypeScript 2.9. Please use the built-in compiler checks instead.`

I'm looking to see why the unused variable isn't being caught and how we can fix that.

KostasRev: Actually it seems the entirety of `tslint` [has been deprecated](https://www.npmjs.com/package/tslint).

test/unit/streaming-client.spec.ts
KostasRev: Discussed and addressed satisfactorily.

test/unit/api-client/api-client.submit.spec.ts
KostasRev: Switch to ESLint proved to be a whole PR's worth of work, so will likely do in a separate issue sometime in the future.

During the attempted move to ESLint it came to my attention that we have several files especially the testing `.js` that do not currently get linted. Additionally while working with ESLint I realized there are a number of options that improve the code quality which we either don't have enabled or might not be working properly due to deprecation. This can all be addressed in a future ticket dedicated to this improvement.

test/unit/api-client/api-client.submit.spec.ts
KostasRev: Addressed this in all async client tests testing for `all options null`.

Note that in [this test](https://github.com/revdotcom/revai-node-sdk/blob/feature/AISDK-92/test/unit/api-client/api-client.submit.spec.ts#L51-L70) the check is also made despite not being in the same pattern as the other null options tests.

src/api-client.ts
```diff
@@ -207,4 +210,12 @@ export class RevAiApiClient {
         return await this.apiHandler.makeApiRequest<Readable>('get',
             url, { 'Accept': contentType || CaptionType.SRT }, 'stream');
     }
+
+    private FilterNullOptions(options: RevAiJobOptions): RevAiJobOptions {
```
amikofalvy: lower case f in filter

examples/async_transcribe_media_from_url.js
```diff
@@ -15,14 +15,17 @@ const token = require('./config/config.json').access_token;
         metadata: "InternalOrderNumber=123456789",
         callback_url: "https://jsonplaceholder.typicode.com/posts",
         skip_diarization: false,
+        skip_punctuation: false,
+        speaker_channels_count: null,
```
amikofalvy: you don't need to include this property in this example.

README.md
```diff
@@ -188,3 +188,17 @@ var customVocabularySubmission = await client.submitCustomVocabularies(customVoc
...
+
+# For Rev.ai NodeSDK Developers
+
+After cloning and installing required npm modules, you should follow these practices when developing:
+
+1. Use the scripts defined in [package.json](https://github.com/revdotcom/revai-node-sdk/tree/develop/package.json) in this manner `npm run [command_name]`:
+    1. `lint` checks that you are not violating any code style standards. This ensures our code's style quality stays high improving readability and reducing room for errors.
+    2. `build` transpiles the typescript into javascript with the options specified in [tsconfig.json](https://github.com/revdotcom/revai-node-sdk/tree/develop/tsconfig.json)
+    3. `unit-test` runs our unit tests which live in the [unit test directory](https://github.com/revdotcom/revai-node-sdk/tree/develop/test/unit).
+    * Note that `integration-test` is currently configured to work with a certain account specified in our continuous integration build environment, as such for now you can check the automated continuous integration checks to pass the integration tests. This will change in the future.
```
amikofalvy: it uses access_token set by environment variable to run the tests. To run this locally, you should set the value as an env var. 

separately,

What do you mean by `This will change in the future?`

README.md
```diff
@@ -188,3 +188,17 @@ var customVocabularySubmission = await client.submitCustomVocabularies(customVoc
...
+2. Add any relevant test logic if you add or modify any features in the source code and check that the tests pass using the scripts mentioned above.
+3. Update the examples provided to illustrate any relevant changes you made, and check that they work properly.
+    * Note that you will have to copy the output of the `build` script into the `examples/node_modules/revai-node-sdk` directory in order to use your changes in the examples. On Unix, this can be simply done with the following command when in the root directory: `$ cp -r dist/src examples/node_modules/revai-node-sdk/`.
```
amikofalvy: This is not a requirement, there are many ways to do this. Mention that this is one method.

amikofalvy: another method is to install from source

amikofalvy: npm install from source

test/unit/api-client/api-client.submit.spec.ts
amikofalvy: I let the mobile team know about the tslint deprecation. You should post about this int he #typescript channel.

amikofalvy (approved):
small changes to the setup docs you wrote.

Post in the slack channel about the tslint deprecation

Fix then :shipit:

README.md
KostasRev: So actually, the tests are sort of hard coded to check for a specific job id that belongs to a specific testing account. @Chanc3 might know more. So while you can configure the environment variable locally, the integration tests are currently only meaningful on the CI environment, where the relevant account is configured. This specific configuration which prevents one from running integration tests locally is what I believe `will change in the future`.

README.md
KostasRev: @amikofalvy how does it look now?

README.md
amikofalvy: Remove the comment about this changing in the future.

PR #42, opened 2020-03-09
AISDK-93: Add remove_disfluencies to the Node SDK

KostasRev:
Branched off AISDK-92, but the diff contains the relevant changes.


amikofalvy (approved):
The implementation looks good. I think we are missing an examples for most job options in the README.md, but since none of the other options have examples we can leave that out of this issue. We should have a discussion with Jenny Chen about better ways to host our documentation.

PR #44, opened 2020-04-17
AISDK-105: Clarfiy mentions of balance in Node SDK

KostasRev:


k-weng (approved):
:shipit:

examples/async_transcribe_local_media_file.js
```diff
@@ -9,7 +9,7 @@ const token = require('./config/config.json').access_token;
     // Get account details
     var account = await client.getAccount();
     console.log(`Account: ${account.email}`);
-    console.log(`Balance: ${account.balance_seconds} seconds`);
+    console.log(`Credits remaining: ${account.balance_seconds} seconds`);
```
k-weng, line 12: Should we change `balance_seconds` to `credits_seconds` or something similar?

KostasRev: I believe not, we only want to change strings and descriptions, not any code.

amikofalvy: Yes. No breaking changes. thanks

PR #45, opened 2020-06-23
 AISDK-109: Add delete endpoint for custom vocabulary

k-weng:


test/integration/test/custom_vocabularies.test.js
```diff
@@ -29,3 +29,17 @@ test('Can retreive submitted custom vocabulary', async (done) => {
         })()
     }, 15000);
 }, 300000);
+
+test('Can delete submitted custom vocabulary', async (done) => {
+    const informationSubmit = await client.submitCustomVocabularies([{phrases:['some','custom','vocabularies']}]);
+
+    var intervalObject = setInterval(function(){
```
amikofalvy, line 36: The integration tests fails because the custom vocabulary did not finish processing. You'll have to wait for the cv to either succeed or fail to delete it.

amikofalvy (approved):
:shipit:  on the implementation
There is one integration test which is failing that needs to be fixed. once it is fixed please merge and assign to QA for final verification.

PR #54, opened 2020-08-24
AISDK-127: Add delete_after_seconds option to streaming and async clients

KostasRev:


test/unit/streaming-client.spec.ts
```diff
@@ -26,7 +26,13 @@ describe('streaming-client', () => {
     describe('start', () => {
         it('Connects to api with parameters', () => {
             // Arrange
-            const config = new SessionConfig('my metadata', 'my custom vocab id', true, true);
+            const config = new SessionConfig(
+                'my metadata', 
+                'my custom vocab id', 
+                true, 
+                true,
+                0
```
k-weng, line 34: Could we change these to named parameters? It will be easier to figure out which arg belongs to which param.

KostasRev: This is sadly not possible.

examples/async_transcribe_local_media_file.js
```diff
@@ -26,7 +26,8 @@ const token = require('./config/config.json').access_token;
             ]
         }],
         filter_profanity: false,
-        remove_disfluencies: false
+        remove_disfluencies: false,
+        delete_after_seconds: 2592000   // 30 days in seconds
```
k-weng:
```suggestion
        delete_after_seconds: 2592000 // 30 days in seconds
```
Very minor - I think too many spaces 😛

k-weng (approved):
:shipit:

PR #55, opened 2020-12-23
AISDK-137: Add language parameter to Node SDK

lrgottlieb:


src/models/streaming/SessionConfig.ts
```diff
@@ -18,18 +19,22 @@ export class SessionConfig {
...
+     *      language code
      */
     constructor(
         metadata?: string,
         customVocabularyID?: string,
         filterProfanity?: boolean,
         removeDisfluencies?: boolean,
-        deleteAfterSeconds?: number
+        deleteAfterSeconds?: number,
+        language?: string
```
amikofalvy, line 31: We don't want to expose the language parameter for streaming. This functionality doesn't exist yet.

PR #56, opened 2020-12-28
AISDK-137: Add language parameter to Node SDK

lrgottlieb:


examples/async_transcribe_local_media_file.js
```diff
@@ -27,7 +27,8 @@ const token = require('./config/config.json').access_token;
         }],
         filter_profanity: false,
         remove_disfluencies: false,
-        delete_after_seconds: 2592000 // 30 days in seconds
+        delete_after_seconds: 2592000, // 30 days in seconds
+        language: false
```
amikofalvy: language is a string type. a boolean value does not make sense here.

amikofalvy (changes requested):
You will need to increase the version number in package.json.

examples/async_transcribe_local_media_file.js
```diff
@@ -27,7 +27,8 @@ const token = require('./config/config.json').access_token;
         }],
         filter_profanity: false,
         remove_disfluencies: false,
-        delete_after_seconds: 2592000 // 30 days in seconds
+        delete_after_seconds: 2592000, // 30 days in seconds
+        language: "en" // ISO 639-1 2-letter language code.
```
amikofalvy, line 31: As we discovered in the manual testing, if someone modifies the example to use a language other than English the API will return a 400 response. I think we should do two things:
1. comment out the extra parameters with explanation:

```
// filter_profanity: false // optional value available with some languages
...
```

2. We can wrap the api execution with a try catch and print the error object to make exception details more clear.

examples/async_transcribe_local_media_file.js
```diff
@@ -15,23 +15,30 @@ const token = require('./config/config.json').access_token;
...
+        filter_profanity: false, //Optional value available with some languages
+        remove_disfluencies: false, //Optional value available with some languages
+        delete_after_seconds: 2592000, // 30 days in seconds
+        language: "en" // ISO 639-1 2-letter language code. 
     };
 
     // Media may be submitted from a local file
-    var job = await client.submitJobLocalFile("./resources/example.mp3", jobOptions);
+    try {
+	    job = await
```
amikofalvy: I don't think this will run correctly. You never declare `job` as var.

examples/async_transcribe_local_media_file.js
```diff
@@ -15,23 +15,30 @@ const token = require('./config/config.json').access_token;
...
+        filter_profanity: false, //Optional value available with some languages
+        remove_disfluencies: false, //Optional value available with some languages
+        delete_after_seconds: 2592000, // 30 days in seconds
+        language: "en" // ISO 639-1 2-letter language code. 
     };
 
     // Media may be submitted from a local file
-    var job = await client.submitJobLocalFile("./resources/example.mp3", jobOptions);
+    try {
+	    var job = await
```
amikofalvy: This won't work either because the job var is now scoped within the try block. The var needs to be declared outside the scope of the try block. I suggest the line above the try block.

```
var job;
try {
...
```

examples/async_transcribe_local_media_file.js
```diff
@@ -15,23 +15,31 @@ const token = require('./config/config.json').access_token;
...
+        remove_disfluencies: false, //Optional value available with some languages
+        delete_after_seconds: 2592000, // 30 days in seconds
+        language: "en" // ISO 639-1 2-letter language code. 
     };
 
     // Media may be submitted from a local file
-    var job = await client.submitJobLocalFile("./resources/example.mp3", jobOptions);
+    var job;
+    try {
+	          job = await
```
amikofalvy: The indent is too strong here. Please use the correct number of tabs or spaces as done in the rest of the file.

examples/async_transcribe_local_media_file.js
```diff
@@ -15,23 +15,31 @@ const token = require('./config/config.json').access_token;
...
+                client.submitJobLocalFile("./resources/example.mp3",
+                jobOptions);
+    } catch (e) {
+            console.dir(e)
```
amikofalvy: indent is too strong here as well

amikofalvy (approved):
Please fix the indentation in the js example then :shipit:

PR #61, opened 2021-01-25
AISDK-146: Streaming client crashes due to write after close

kbridbur:


src/streaming-client.ts
```diff
@@ -158,6 +163,7 @@ export class RevAiStreamingClient extends EventEmitter {
 
     private closeStreams(): void {
         this.requests.end();
+        this.streamsClosed = true;
```
kbridbur, line 166: Moved this line from where it was in the suggestion left by Steve. 

There is a still a race here where a message can come in just before this line but not get written to the duplex until after the null is pushed but I think by changing the order of these lines we reduce the likelyhood of this race

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "revai-node-sdk",
-  "version": "2.6.0",
+  "version": "2.7.0",
```
amikofalvy: I would suggest it is a patch fix for 2.6.1

amikofalvy (approved):

kbridbur:
One of the integration tests appears to be fragile and failing but that is outside the scope of this issue. I have filed another issue to fix the fragile integration test

PR #62, opened 2021-02-05
AISDK-151: Properly check for stream state before reading/writing

kbridbur:


src/models/streaming/BufferedDuplex.ts
```diff
@@ -17,37 +16,35 @@ export class BufferedDuplex extends Duplex {
         super(options);
         this.input = input;
         this.output = output;
-        this.areOutputHandlersSetup = false;
+        this.setupInput();
```
kbridbur, line 19: These handlers were the main thing we were missing before. They ensure that events from the inner streams are propagated to users of our BufferedDuplex

test/unit/streaming-client.spec.ts
```diff
@@ -249,6 +253,7 @@ describe('streaming-client', () => {
 
             // Assert
             expect(() => { duplex.write("message"); }).toThrow();
+            expect(duplex.writable).toBe(false);
```
kbridbur, line 256: test currently fails on this because the duplex doesn't emit the `end` event until the input stream drains the final EOS message and the stream isn't marked as unwritable until the `end` event fires. Working on finding a way to make the client's end function wait until either the `end` event or the `drain` event are fired before returning

src/streaming-client.ts
```diff
@@ -95,6 +95,7 @@ export class RevAiStreamingClient extends EventEmitter {
      * Signals to the api that you have finished sending data.
      */
     public end(): void {
+        this.requests.emit('ending');
```
kbridbur, line 98: This event allows us to sync the ending of the request stream with the ending of the overlying duplex stream. Without this there is a period of time where the requests stream is draining the final input message and is not writable but has not yet emitted the 'finish' event which we would use to notify the duplex that it should finish as well. This way we signal the duplex to set writable = false and stop accepting new input while the underlying request stream finishes inserting the last EOS message

test/unit/streaming-client.spec.ts
```diff
@@ -133,6 +134,7 @@ describe('streaming-client', () => {
 
             // Assert
             expect(connectionError).toBe(expectedError);
+            expect(res.writable).toBe(false);
```
amikofalvy, line 141: Should we add an assertion to at one other valid test case for when `expect(res.writable).toBe(true);` ?

amikofalvy (approved):
The code changes look good. I suggest we add a test to make sure there are no regressions on the fix for the behavior described by the customer.

amikofalvy:
Also, the integration tests are failing for this branch's travis build, but I think these are spurious failures. The integration tests pass for me locally.

PR #76, opened 2022-01-19
AISDK-166: Prevent bytes in form of EOS in data from ending prematurely

k-weng:
Things done in this PR
- [x] Fix original issue where bytes in the form of EOS can lead to SDK prematurely sending a EOS UTF-8 message to Rev.ai Streaming API
- [x] Upgrade lock file to version 2
- [x] Added typings for module `websocket` for better static typing

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,12 +35,13 @@
   "dependencies": {
     "axios": "^0.21.1",
     "form-data": "^2.3.3",
-    "websocket": "^1.0.28"
+    "websocket": "^1.0.34"
   },
   "devDependencies": {
     "@types/form-data": "^2.2.1",
     "@types/jest": "^23.3.13",
     "@types/node": "^10.12.18",
+    "@types/websocket": "^1.0.4",
     "dotenv": "^8.1.0",
     "jest": "^24.8.0",
     "ts-jest": "^24.0.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "revai-node-sdk",
-  "version": "2.6.3",
+  "version": "3.0.0",
   "description": "Rev.ai makes speech applications easy to build!",
   "main": "src/index.js",
   "scripts": {
--- a/src/models/streaming/BufferedDuplex.ts
+++ b/src/models/streaming/BufferedDuplex.ts
@@ -1,4 +1,4 @@
-import { Duplex, PassThrough } from 'stream';
+import { Duplex, DuplexOptions, PassThrough } from 'stream';
 
 /**
  * Represents a two way buffered stream. Extends Duplex which implements both Readable and Writable
@@ -12,7 +12,7 @@ export class BufferedDuplex extends Duplex {
      * @param output Buffer for the Readable side of the stream.
      * @param options Options to be passed through to the superclass.
      */
-    constructor (input: PassThrough, output: PassThrough, options?: any) {
+    constructor (input: PassThrough, output: PassThrough, options?: DuplexOptions) {
         super(options);
         this.input = input;
         this.output = output;
@@ -20,7 +20,7 @@ export class BufferedDuplex extends Duplex {
         this.setupOutput();
     }
 
-    public _write(chunk: any, encoding: string, callback: any): boolean {
+    public _write(chunk: any, encoding: string, callback: (error?: Error | null) => void): boolean {
         const needsDrain = this.input.write(chunk, encoding, () => needsDrain && callback());
         if (!needsDrain) {
             this.input.once('drain', callback);
@@ -28,12 +28,12 @@ export class BufferedDuplex extends Duplex {
         return needsDrain;
     }
 
-    public _read(size: number): any {
+    public _read(size: number): void {
         const chunk = this.output.read(size);
         if (chunk !== null) {
             this.push(chunk);
         } else {
-            this.output.once('readable', s => this._read(s));
+            this.output.once('readable', (s: number) => this._read(s));
         }
     }
 
--- a/src/models/streaming/BufferedDuplex.ts
+++ b/src/models/streaming/BufferedDuplex.ts
@@ -39,7 +39,6 @@ export class BufferedDuplex extends Duplex {
 
     private setupInput(): void {
         this.once('finish', () => this.input.end());
-        this.input.on('ending', () => this.writable = false);
         this.input.on('finish', () => this.end());
         this.input.on('error', error => this.emit('error', error));
     }
--- a/src/streaming-client.ts
+++ b/src/streaming-client.ts
@@ -36,6 +36,7 @@ export class RevAiStreamingClient extends EventEmitter {
     private config: AudioConfig;
     private requests: PassThrough;
     private responses: PassThrough;
+    private protocol: Duplex;
 
     /**
      * @param accessToken Access token associated with the user's account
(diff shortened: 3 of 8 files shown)


src/streaming-client.ts
```diff
@@ -49,7 +50,11 @@ export class RevAiStreamingClient extends EventEmitter {
         this.baseUrl = `wss://api.rev.ai/speechtotext/${version}/stream`;
         this.requests = new PassThrough();
         this.responses = new PassThrough({ objectMode: true });
-        this.client = new client({ keepalive: true, keepaliveInterval: 30000 });
+        this.client = new client({
+            // @ts-ignore
+            keepalive: true,
+            keepaliveInterval: 30000
+        });
```
k-weng, line 62: Needed to add `@ts-ignore` because the typings for `ClientConfig` for the `websocket` package does not include `keepalive` or `keepaliveInterval` but they are still used - https://github.com/theturtle32/WebSocket-Node/blob/574b703ca978709832c6c2f67d6c1de376d432b5/lib/WebSocketConnection.js#L111

src/streaming-client.ts
```diff
@@ -88,7 +93,10 @@ export class RevAiStreamingClient extends EventEmitter {
         }
 
         this.client.connect(url);
-        return new BufferedDuplex(this.requests, this.responses, { readableObjectMode: true });
+        return new BufferedDuplex(this.requests, this.responses, {
+            readableObjectMode: true,
+            writableObjectMode: true
```
k-weng: This is part of the proposed fix, enabling `writableObjectMode`. If "EOS" is written to the stream, when reading the chunk, it would return a buffer of length 3. If "EOS" bytes were present in the binary data from media, the read chunk would be longer than 3.

src/streaming-client.ts
```diff
@@ -145,20 +153,21 @@ export class RevAiStreamingClient extends EventEmitter {
...
+        if (conn.connected) {
             let value = buffer.read(buffer.readableLength);
             if (value !== null) {
-                connection.send(value);
-                if (value.includes('EOS') || value.includes(Buffer.from('EOS'))) {
-                    connection.sendUTF('EOS');
+                if (Buffer.compare(value, Buffer.from('EOS')) === 0) {
+                    conn.sendUTF('EOS');
+                } else {
+                    conn.send(value);
```
k-weng: If read buffer is just EOS, then we'll send EOS. If not, we send the full buffer.

test/unit/testhelpers.ts
```diff
@@ -1,4 +1,4 @@
-import { AxiosError } from "axios";
```
k-weng, line 1: Changes in here were fixes for `tslint`

test/unit/mocks/websocket-mock.ts
```diff
@@ -4,7 +4,7 @@ export class WebSocketConnectionMock extends EventEmitter {
     connected: boolean;
```
k-weng, line 4: Changes in here were fixes for `tslint`

src/streaming-client.ts
kbridbur: This was the state of the sdk some time ago. We need to be careful here because we're reading buffer.readableLength. this means if there are multiple messages written to the buffer well read all of them at once. This why we had to check for contains(EOS) and not just equals(EOS). Does enabling writableObjectMode somehow change the way buffer.readableLength works so we're guaranteed to only get one chunk at a time?

k-weng: @kbridbur I actually enabled `objectMode` to the wrong stream - I needed to enable it in the underlying `requests` stream - here's the fix commit - https://github.com/revdotcom/revai-node-sdk/pull/76/commits/c3c9b0593808d0ef05eccecf0d50f8d11716da82 - now when reading, if EOS was provided as a string, it will be read as a string, therefore we can check typing and content when reading.

To answer your question about reading and object mode. If in object mode, read will always return a single object no matter the `readableLength` (which in this case is the number of objects now instead of number of bytes) - [source](https://nodejs.org/api/stream.html#:~:text=always%20return%20a%20single%20item)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "revai-node-sdk",
-  "version": "2.6.3",
+  "version": "3.0.0",
```
k-weng, line 3: @amikofalvy @kbridbur I pushed it to a major version since the change would affect customers who may have been submitting to the stream objects other than buffers (they shouldn't be but just in case)

src/streaming-client.ts
```diff
@@ -145,20 +152,21 @@ export class RevAiStreamingClient extends EventEmitter {
...
             });
-            this.doSendLoop(connection, this.requests);
+            this.doSendLoop(conn, this.requests);
         });
     }
 
-    private doSendLoop(connection: any, buffer: PassThrough): void {
-        if (connection.connected) {
-            let value = buffer.read(buffer.readableLength);
+    private doSendLoop(conn: connection, buffer: PassThrough): void {
```
kbridbur: I wonder why they made the Message type capital but connection isn't :(

src/streaming-client.ts
```diff
@@ -145,20 +152,21 @@ export class RevAiStreamingClient extends EventEmitter {
...
+        if (conn.connected) {
+            const value = buffer.read();
             if (value !== null) {
-                connection.send(value);
-                if (value.includes('EOS') || value.includes(Buffer.from('EOS'))) {
-                    connection.sendUTF('EOS');
+                if (typeof value === 'string' && value === 'EOS') {
```
kbridbur: Should we simplify this? It could be
`if (typeof value === 'string') { conn.sendUTF(value) }`

This would make the SDK more flexible but would make it so customers can mess up the protocol more easily

k-weng: Yeah I'm fine with this

kbridbur (approved):
lgtm

src/models/streaming/BufferedDuplex.ts
```diff
@@ -12,28 +12,28 @@ export class BufferedDuplex extends Duplex {
...
+    constructor (input: PassThrough, output: PassThrough, options?: DuplexOptions) {
         super(options);
         this.input = input;
         this.output = output;
         this.setupInput();
         this.setupOutput();
     }
 
-    public _write(chunk: any, encoding: string, callback: any): boolean {
+    public _write(chunk: any, encoding: string, callback: (error?: Error | null) => void): boolean {
```
dmtrrk, line 23: should type `error?: Error | null` lead to `Error | null | undefined`?
and what is the purpose of nullable error? is it for backward compatibility?

src/streaming-client.ts
```diff
@@ -47,9 +48,13 @@ export class RevAiStreamingClient extends EventEmitter {
         this.accessToken = accessToken;
         this.config = config;
         this.baseUrl = `wss://api.rev.ai/speechtotext/${version}/stream`;
-        this.requests = new PassThrough();
+        this.requests = new PassThrough({ objectMode: true });
```
dmtrrk, line 52: in object mode the buffering works in a [different ](https://nodejs.org/api/stream.html#object-mode) way. so the buffering threshold works in a different way: `highWaterMark` option param sets the count obj objects to buffer rather than the buffer size in bytes.
And [I don't](https://github.com/nodejs/node/blob/master/lib/internal/streams/state.js#L11) see the default buffering value by default (it looks like it is 16 objects by default).
Is it correct default setup?

src/models/streaming/BufferedDuplex.ts
k-weng: Yeah this is the same as the `Duplex`'s typings in `@types/node`
```
// Note: Duplex extends both Readable and Writable.
class Duplex extends Readable implements Writable {
    writable: boolean;
    readonly writableHighWaterMark: number;
    readonly writableLength: number;
    constructor(opts?: DuplexOptions);
    _write(chunk: any, encoding: string, callback: (error?: Error | null) => void): void;
```

src/streaming-client.ts
k-weng: I think default is fine

PR #78, opened 2022-01-27
AISDK-155: Add support for human transcription

k-weng:
Add support for human transcription option - To make things simpler, I added submission of human transcription jobs as separate functions from the normal machine transcription job submissions.

Tasks
- [x] Implementation
- [x] Tests

### DIFF
--- a/src/api-client.ts
+++ b/src/api-client.ts
@@ -5,7 +5,7 @@ import { Readable } from 'stream';
 import { ApiRequestHandler } from './api-request-handler';
 import { CaptionType } from './models/async/CaptionType';
 import { RevAiAccount } from './models/async/RevAiAccount';
-import { RevAiJobOptions } from './models/async/RevAiJobOptions';
+import { RevAiHumanTranscriptionJobOptions, RevAiJobOptions } from './models/async/RevAiJobOptions';
 import { RevAiApiJob } from './models/RevAiApiJob';
 import { RevAiApiTranscript } from './models/RevAiApiTranscript';
 
@@ -59,7 +59,7 @@ export class RevAiApiClient {
      * @returns List of job details
      */
     async getListOfJobs(limit?: number, startingAfter?: string): Promise<RevAiApiJob[]> {
-        let params = [];
+        const params = [];
         if (limit) {
             params.push(`limit=${limit}`);
         }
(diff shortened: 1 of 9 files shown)


src/api-client.ts
```diff
@@ -147,6 +147,79 @@ export class RevAiApiClient {
...
+     * @returns Details of the submitted job
+     */
+    async submitHumanTranscriptionJobUrl(
+        mediaUrl: string,
+        options?: RevAiHumanTranscriptionJobOptions
+    ): Promise<RevAiApiJob> {
+        options = this.filterNullOptions({
+            transcriber: 'human',
+            media_url: mediaUrl,
+            ...(options || {})
```
kbridbur: what if they also specify transcriber in these options? If we're going to have separate methods for this then `transcriber` should go into the `RevAiJobOptions` subclass

src/api-client.ts
```diff
@@ -147,6 +147,79 @@ export class RevAiApiClient {
...
+
+    /**
+     * See https://www.rev.ai/docs#section/Human-Transcription-(Labs)
+     * Send local file for transcription by a human.
+     * @param filepath Path to local file to be transcribed. Assumes the process has access to read this file.
+     * @param options (optional) Options submitted with the job, see RevAiHumanTcJobOptions object
+     *     or https://www.rev.ai/docs#operation/SubmitTranscriptionJob
+     * @returns Details of submitted job
+     */
+    async submitHumanTranscriptionJobLocalFile(
```
kbridbur: All these methods need testing

kbridbur (changes requested):
transcriber should be moved and tests should be added

src/api-client.ts
amikofalvy: I understand the motivation for creating a separate transcription function for human, but I have the same worry as Kyle that this introduces additional complications.

One of the selling points of having human and machine in the same transcription api is ease of implementation and switch. I believe that changing a single option `transcriber` is a simpler change than to replace the name of a function.

If you were writing conditionals which set the transcriber you would have to write more if you had to use another function:

```
if (audioQuality < 0.7) {
    submitHumanTranscriptionJobUrl(...)
} else {
    submitJobUrl(...)
}
```

In the other way:

```
submitJobUrl(..., {...options, transcriber: audioQuality < 0.7 ? 'human' : 'machine_v2' );
```

amikofalvy (approved):

kbridbur (approved):

PR #79, opened 2022-02-02
AISDK-170: Support Node 14, 16 and 17

k-weng:
Checking to make sure our integration tests work with Node 14, 16 and 17. Also locally ran the streaming example for the three latest versions.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -2,7 +2,7 @@
 
 name: CI
 
-# Controls when the action will run. 
+# Controls when the action will run.
 on:
   # Triggers the workflow on push or pull request events but only for the develop branch
   push:
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x]
+        node-version: [8.x, 10.x, 12.x, 14.x, 16.x, 17.x]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x, 16.x, 17.x]
+        node-version: [8.x, 10.x, 12.x, 14.x, 16.x, 17.x]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x, 17.x]
+        node-version: [10.x, 12.x, 14.x, 16.x, 17.x]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x, 14.x, 16.x, 17.x]
+        node-version: [12.x, 14.x, 16.x, 17.x]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/build_test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x]
+        node-version: [8.x, 10.x, 12.x, 14.x, 16.x, 17.x]
```
k-weng, line 24: @amikofalvy When should we stop supporting `8` and `10`?

README.md
```diff
@@ -58,7 +58,6 @@ const job = await client.submitJobAudioData(stream, "file.mp3");
 
 You can also submit a job to be handled by a human transcriber using our [Human Transcription](https://www.rev.ai/docs#section/Human-Transcription-(Labs)) option.
 ```javascript
-// Submitting with human transcription options
```
k-weng, line 61: Slight redundant comment by me

test/unit/streaming-client.spec.ts
```diff
@@ -113,42 +113,40 @@ describe('streaming-client', () => {
...
 
             // Act
             mockClient.emit('httpResponse', { statusCode: 401 });
 
             // Assert
             expect(statusCode).toBe(401);
             expect(res.writable).toBe(false);
-            expect(() => { res.write('message'); }).toThrow();
+            res.on('error', (err: any) => expect(err.code).toBe('ERR_STREAM_WRITE_AFTER_END'));
+            res.write('messsage');
```
k-weng: The original fails unit tests for the new Node versions. It's because `res` doesn't actually throw but emits an error because it extends from `EventEmitter`, so to handle the end of stream error, we provide a listener and add an assertion in there.

.github/workflows/build_test.yml
kbridbur: IMO no, we can continue to support them until it makes our lives more difficult. As of right now that code that supports 14.x, 16.x, and 17x inherently supports 8 and 10. If that ever stops being the case we can drop them

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "src/index.js",
   "scripts": {
     "build": "tsc",
+    "build-examples": "tsc && cd ./examples && npm install && rm -r ./node_modules/revai-node-sdk/src && cp -r ../dist/src ./node_modules/revai-node-sdk",
```
kbridbur, line 8: The `cd ./examples` is a little suspicious. It will actually change where to user's terminal is focused, no?

src/streaming-client.ts
```diff
@@ -179,8 +179,9 @@ export class RevAiStreamingClient extends EventEmitter {
     }
 
     private closeStreams(): void {
-        if (this.streamsClosed === false) {
+        if (!this.streamsClosed) {
             this.streamsClosed = true;
+            this.protocol.end();
```
kbridbur, line 184: This shouldn't be necessary. According to the node 17 docs:
```
duplex.allowHalfOpen (https://nodejs.org/api/stream.html#duplexallowhalfopen)
If false then the stream will automatically end the writable side when the readable side ends. Set initially by the allowHalfOpen constructor option, which defaults to false.
```
Ending requests should end the protocol duplex on its own

kbridbur: I guess I don't see this in the node 14 docs. Is it necessary for that version?

k-weng: I think the `requests` and `responses` Passthroughs is separate from the duplex readable and writable streams? Several unit-tests started failing because the assertion `expect(res.writable).toBe(false)` was failing (where `res` is the returned duplex)

package.json
k-weng: Might be wrong but I believe the npm script can only be run from the project root directory (where `package.json` lives) - so running `npm run build-examples` -> `cd ./examples` will be in respect to the project root directory - if it was `cd examples` it may be more suspicious?

.github/workflows/build_test.yml
amikofalvy: I agree with Kyle

amikofalvy:
It looks like we are missing integration test running for node 17

amikofalvy (approved):

PR #82, opened 2022-02-04
AISDK-141: Support custom_vocabulary_id for async client

k-weng:


examples/async_transcribe_local_media_file.js
```diff
@@ -17,6 +17,7 @@ const token = require('./config/config.json').access_token;
         skip_diarization: false,
         skip_punctuation: false, // Optional value available with some languages
         speaker_channels_count: null, // Optional value available with some languages
+        custom_vocabulary_id: null, // Optional vaule available with some languages
```
beaudrychase:
```suggestion
        custom_vocabulary_id: null, // Optional value available with some languages
```

beaudrychase (approved):

beaudrychase (commented):
Actually aren't there tests that need to have this option added? like this one
https://github.com/revdotcom/revai-node-sdk/blob/1356858172e046e24f40c0a10259a75f53fd50b3/test/unit/api-client/api-client.submit.spec.ts#L83

PR #91, opened 2022-03-30
AISDK-176: Update rev.ai references to Rev AI

amikofalvy:


kshiflett88 (approved):
When looking through there is one in src/LICENSE not sure if that needs to be changed as well https://github.com/revdotcom/revai-node-sdk/blob/f5aca04108f430432be23a47546b42d6ca2d2b58/src/LICENSE#L3 but other than that I think you got them all. So after making that last change (if needed) 🚢

amikofalvy:
I made a note to check with legal regarding the year of the copyright on our LICENSE.

PR #92, opened 2022-04-01
AISDK-183: Implement Link Checker in Node SDK

kshiflett88:
Implemented based off of https://github.com/revdotcom/revai-java-sdk/pull/37

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,6 +1,6 @@
 # API Domain stubs used as url partials in the api clients
 api.rev.ai
-\${configHelper\.getBaseUrl\(\)}
+speechtotext\/v1\/
 
 # example domains used in unit tests
 www.example.com
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,5 +1,8 @@
 # API Domain stubs used as url partials in the api clients
 api.rev.ai
+https:// + ${configHelper.getBaseUrl()} + /speechtotext/v1/
+wss:// + ${configHelper.getBaseUrl()} + /speechtotext/v1/stream
+https:// + ${configHelper.getBaseUrl()} + /speechtotext/v1/vocabularies
 
 # example domains used in unit tests
 www.example.com
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,8 +1,8 @@
 # API Domain stubs used as url partials in the api clients
 api.rev.ai
-https:// + ${configHelper.getBaseUrl()} + /speechtotext/v1/
-wss:// + ${configHelper.getBaseUrl()} + /speechtotext/v1/stream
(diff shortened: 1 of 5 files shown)


test/unit/api-client/api-client.general.spec.ts
```diff
@@ -25,7 +25,7 @@ describe('api-client', () => {
 
     describe('getAccount', () => {
         it('get account email and balance', async () => {
-            const accountEmail = 'test@rev.com';
+            const accountEmail = 'example.com';
```
k-weng: This doesn't make sense to change to a link just to satisfy link checker since `accountEmail` should be an email - why not just ignore emails in link checker?

test/unit/testhelpers.ts
```diff
@@ -43,7 +43,7 @@ export const setupFakeApiError = (
 export const setupFakeInvalidStateError = (): AxiosError => {
     let err = setupFakeApiError(409,
         'Job is in invalid state',
-        'https://www/rev.ai/api/speech/v1/errors/invalid-job-state',
+        'https://www.example.com/api/speech/v1/errors/invalid-job-state',
```
k-weng: Instead of changing these to `example.com` to satisfy link checker ignores - should fix to point to the actual error code links - which is at `https://www.rev.ai/api/v1/errors/<error-code>` so here it would be
```suggestion
        'https://www.rev.ai/api/v1/errors/invalid-job-state',
```

Please fix for the others

.lycheeignore
```diff
@@ -0,0 +1,6 @@
+# API Domain stubs used as url partials in the api clients
+api.rev.ai
+getBaseUrl
```
k-weng: Checking your failed build logs - https://github.com/revdotcom/revai-node-sdk/runs/5796182915?check_suite_focus=true#step:4:73- should this be lowercased to work?
```suggestion
getbaseurl
```

k-weng (changes requested):
Couple of comments

test/unit/testhelpers.ts
kshiflett88: @k-weng these redirect to `https://docs.rev.ai/api/asynchronous/#error-codes` is that meant to happen or should these links be to a page that shows this specific error and one you can see if you check the console/network when inspecting?

k-weng: It's expected - that's where all of our error codes are/will be documented

k-weng (approved):

test/unit/api-client/api-client.general.spec.ts
amikofalvy: We should be using the example.com domain name, but we could have made it an email like `test@example.com`

PR #100, opened 2022-05-12
AISDK-200: Add topic extraction client

hbkse:


examples/topic_extraction_example.js
```diff
@@ -0,0 +1,9 @@
+const revai = require('revai-node-sdk');
+const token = require('./config/config.json').access_token;
+
+(async () => {
+    // Initialize your client with your Rev AI access token
+    var client = new revai.TopicExtractionClient(token);
```
dmtrrk:
```suggestion
    const client = new revai.TopicExtractionClient(token);
```

examples/topic_extraction_example.js
```diff
@@ -0,0 +1,9 @@
...
+
+
```
dmtrrk:
```suggestion
   // put your logic with the Topic Extraction here
```

src/base-api-client.ts
```diff
@@ -0,0 +1,96 @@
+import { ApiRequestHandler } from './api-request-handler';
+
+export type ServiceApi = 'speechtotext' | 'topic_extraction' | 'sentiment_analysis';
+
+/**
+ *
```
dmtrrk: Empty summary?

src/base-api-client.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * 
+     * @param mediaUrl Web location of media to be downloaded and transcribed
+     * @param options (optional) Options submitted with the job: see RevAiJobOptions object
+     * @returns Details of the submitted job
+     */
+    protected async _submitJob<TJob>(options?: any): Promise<TJob> {
+        options = this.filterNullOptions({
```
dmtrrk: can't we do this?

```suggestion
        options = this.filterNullOptions(options || {})?
```

src/base-api-client.ts
```diff
@@ -0,0 +1,96 @@
...
+            ...(options || {})
+        });
+
+        return await this.apiHandler.makeApiRequest<TJob>('post', `/jobs`,
+            { 'Content-Type': 'application/json' }, 'json', options);
+    }
+
+    /**
+     * 
+     * @param params
```
dmtrrk: please add some summary if it is needed for protected methods

src/base-api-client.ts
```diff
@@ -0,0 +1,96 @@
...
+        return await this.apiHandler.makeApiRequest<TJob>('post', `/jobs`,
+            { 'Content-Type': 'application/json' }, 'json', options);
+    }
+
+    /**
+     * 
+     * @param params 
+     * @returns 
+     */
+     protected async _getResult<TResult>(id: string, params?: any): Promise<TResult> {
```
dmtrrk: can we have a type defined in this file e.g.
```
type ParamsType = [key: string]: string;
```
and then have it something like this:
```
 protected async getResult<TResult>(id: string, params?: ParamsType): Promise<TResult> {
```

src/models/RevAiApiJob.ts
```diff
@@ -2,6 +2,9 @@ import { SegmentToTranscribe } from './async/RevAiJobOptions';
 import { JobStatus } from './JobStatus';
 import { JobType } from './JobType';
 
+/**
+ *
```
dmtrrk: add a summary

src/models/topic-extraction/TopicExtractionJob.ts
```diff
@@ -0,0 +1,19 @@
+import { JobStatus } from '../JobStatus';
+
+/**
+ *
```
dmtrrk: summary

src/models/topic-extraction/TopicExtractionResult.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ *
```
dmtrrk: summary

src/models/topic-extraction/TopicExtractionJobOptions.ts
```diff
@@ -0,0 +1,14 @@
+import { RevAiApiTranscript } from "../RevAiApiTranscript";
+
+/**
+ * Options that can used when submitting Rev AI topic extraction job.
+ * See https://docs.rev.ai/api/topic-extraction/reference/#operation/SubmitTopicExtractionJob for more details.
+ */
+export interface TopicExtractionJobOptions {
+    metadata?: string;
+    callbackUrl?: string;
```
dmtrrk: are you sure we are getting camelCase rather than snake case responses?

src/models/topic-extraction/TopicExtractionJob.ts
```diff
@@ -0,0 +1,19 @@
+import { JobStatus } from '../JobStatus';
+
+/**
+ * 
+ */
+export interface TopicExtractionJob {
+    type: 'topic_extraction';
```
dmtrrk: why do we need this field here?
In UI we used it for the DU purpose to distinguish this job type from others

src/topic-extraction-client.ts
```diff
@@ -0,0 +1,44 @@
...
+export class TopicExtractionClient extends BaseApiClient {
+    /**
+     * @param accessToken Access token used to validate API requests
+     * @param version (optional) version of the API to be used
+     */
+    constructor (accessToken: string, version = 'v1') {
+        super(accessToken, 'topic_extraction', version);
+    }
+
+    async getJobDetails(id: string): Promise<TopicExtractionJob> {
```
dmtrrk, line 29: I think we need summaries for public methods

dmtrrk (changes requested):

src/models/topic-extraction/TopicExtractionJobOptions.ts
hbkse: yeah, these should be snake case

src/topic-extraction-client.ts
```diff
@@ -0,0 +1,24 @@
...
+import { TopicExtractionJobOptions } from './models/topic-extraction/TopicExtractionJobOptions';
+import { TopicExtractionResult } from './models/topic-extraction/TopicExtractionResult';
+import { TopicExtractionResultOptions } from './models/topic-extraction/TopicExtractionResultOptions';
+
+const TopicsContentType = 'application/vnd.rev.topic.v1.0+json';
+
+/**
+ * Client which handles connection to the Rev AI topic extraction API.
+ */
+export class TopicExtractionClient extends BaseApiClient<TopicExtractionJob, TopicExtractionJobOptions, TopicExtractionResult, TopicExtractionResultOptions> {
```
hbkse: does anyone have a suggestion for how to split this into multiple lines?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "revai-node-sdk",
-  "version": "3.2.0",
+  "version": "3.3.0",
```
kbridbur: If imports have potentially been changed this needs a major version bump

src/base-api-client.ts
```diff
@@ -0,0 +1,99 @@
...
+            { 'Content-Type': 'application/json' }, 'json', options);
+    }
+
+    /**
+     * Get the result of a job. 
+     * @param id id of job to get result of
+     * @param options (optional) Options submitted with the request
+     * @returns Job result object
+     */
+    protected async getResult(id: string, options?: TResultOptions, headers: any = {}): Promise<TResult> {
```
kbridbur: Does headers need to be any? I see it as `{}` [here](https://github.com/revdotcom/revai-node-sdk/blob/develop/src/api-request-handler.ts#L39)

kbridbur: I'm not sure I see the value in TResultOptions over a `{}` you're essentially just using the object as a dictionary

src/base-api-client.ts
```diff
@@ -0,0 +1,99 @@
...
+        options = this.filterNullOptions(options || {});
+        const query = this.buildQueryParams(options || {});
+        return await this.apiHandler.makeApiRequest<TResult>('get',
+            `/jobs/${id}/result${query ? `?${query}` : ''}`, headers, 'json');
+    }
+
+    private buildQueryParams(params: any): string {
```
kbridbur, line 86: TResultOptions

kbridbur:
```suggestion
    private buildQueryParams(params: {}): string {
```

src/base-api-client.ts
```diff
@@ -0,0 +1,99 @@
...
+        return Object.keys(params).map((key) => `${key}=${params[key]}`).join('&');
+    }
+
+    private filterNullOptions(options: any): any {
```
kbridbur:
```suggestion
    private filterNullOptions(options: TResultOptions): {} {
```

src/base-api-client.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Get a list of jobs submitted within the last 30 days in reverse chronological order 
+     * (last submitted first) up to the provided limit number of jobs per call. Pagination is supported via passing
+     * the last job id from previous call into starting_after.
+     * @param limit (optional) maximum number of jobs to retrieve, default is 100
+     * @param startingAfter (optional) returns only jobs created after the job with this id, exclusive
+     * @returns List of job details
+     */
+    protected async getListOfJobs(limit?: number, startingAfter?: string): Promise<TJob[]> {
```
kbridbur: Another reason I'm not a fan of TResultOptions. It makes our interfaces inconsistent with this method. In some cases the options are nullable parameters, in others they're objects

src/models/JobStatus.ts
```diff
@@ -1,5 +1,6 @@
 export enum JobStatus {
     Failed = 'failed',
     InProgress = 'in_progress',
-    Transcribed = 'transcribed'
+    Transcribed = 'transcribed',
+    Completed = 'completed'
```
kbridbur, line 5: Should we consider creating a new enum for jobs statuses? Only old job types use `transcribed`

src/models/topic-extraction/TopicExtractionJob.ts
```diff
@@ -0,0 +1,17 @@
+import { JobStatus } from '../JobStatus';
+
+/** Topic extraction job model */
+export interface TopicExtractionJob {
+    id: string;
+    status: JobStatus;
+    type: string;
```
kbridbur:
```suggestion
    type: JobType;
```
Add topic and sentiment to the job type enum

src/topic-extraction-client.ts
```diff
@@ -0,0 +1,24 @@
...
+    /**
+     * @param accessToken Access token used to validate API requests
+     * @param version (optional) version of the API to be used
+     */
+    constructor (accessToken: string, version = 'v1') {
+        super(accessToken, 'topic_extraction', version);
+    }
+
+    async getResult(id: string, params?: TopicExtractionResultOptions) {
+        return super.getResult(id, params, { 'Accept': TopicsContentType });
```
kbridbur: One concern is that user's could accidentally call the `super` method very easily by accident. Not sure how we want to address this

hbkse: I guess this is a reason why we should have all the base methods with underscores and don't let the user invoke directly

src/models/JobStatus.ts
hbkse: I think we would just keep using the same enum but mark the `transcribed` field as deprecated eventually

src/base-api-client.ts
hbkse: at the time it was so that we could enforce a type since users would invoke this method directly but I don't think it's relevant anymore

hbkse: following that train of thought, there's probably no reason to have a TSubmitOptions on this base api client

package.json
hbkse: I don't think any existing imports have been changed

test/unit/base-api-client.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+            makeApiRequest: mockMakeApiRequest
+        }));
+        sut = new BaseApiClient('testtoken', 'topic_extraction');
+    });
+
+    describe('_getJobDetails', () => {
+        it('get job by id', async () => {
+            mockMakeApiRequest.mockResolvedValue(jobDetails);
+
+            const job = await (sut as any)._getJobDetails(jobId);
```
hbkse: Since these are protected methods, we do this cast to `any`. Open to feedback on this

src/models/JobStatus.ts
hbkse: I guess one thing is that we have this CustomVocabularyStatus enum, but I don't think it makes sense for us to define new enums for every single service type

examples/topic_extraction_example.js
```diff
@@ -0,0 +1,52 @@
...
+    /** Submit a job with whatever text you want by changing this input */
+    const text = "An umbrella or parasol is a folding canopy supported by wooden or metal ribs that is  \
+        usually mounted on a wooden, metal, or plastic pole. It is designed to protect a person \
+        against rain or sunlight. The term umbrella is traditionally used when protecting oneself from \
+        rain, with parasol used when protecting oneself from sunlight, though the terms continue to be \
+        used interchangeably. Often the difference is the material used for the canopy; some parasols \
+        are not waterproof, and some umbrellas are transparent. Umbrella canopies may be made of \
+        fabric or flexible plastic. There are also combinations of parasol and umbrella that are \
+        called en-tout-cas (French for 'in any case').";
+    job = await client.submitJobFromText(text, jobOptions);
```
kbridbur:
```suggestion
    let job = await client.submitJobFromText(text, jobOptions);
```
We can also change the one below. I think its fair to expect one or the other to be commented and it makes the file more readable

examples/topic_extraction_example.js
```diff
@@ -0,0 +1,52 @@
...
+    // job = await client.submitJobFromTranscript(transcript, jobOptions);
+
+    console.log('Topic extraction job submitted.');
+    console.log(`Job Id: ${job.id}`);
+    console.log('Polling for job completion...');
+
+    while((jobStatus = (await client.getJobDetails(job.id)).status) == revai.JobStatus.InProgress)
+    {
+        console.log(`Job ${job.id} is ${jobStatus}`);
+        await new Promise( resolve => setTimeout(resolve, 5000));
```
kbridbur:
```suggestion
        await new Promise( resolve => setTimeout(resolve, 1000));
```
Topic jobs finish faster than async so this can be lower

examples/topic_extraction_example.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    const topics = await client.getResult(job.id);
+    console.log(JSON.stringify(topics))
```
kbridbur: Can you post an example of what this looks like? Is it readable for a user?

examples/topic_extraction_example.js
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Delete a job.
+     * Job deletion will remove all information about the job from the servers
+     */
+    await client.deleteJob(job.id);
```
kbridbur: Lets comment this out by default 
```suggestion
    // await client.deleteJob(job.id);
```

src/base-api-client.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export class BaseApiClient<TJob, TResult> {
+    apiHandler: ApiRequestHandler;
+
+    /**
+     * @param accessToken Access token used to validate API requests
+     * @param serviceApi Type of api service
+     * @param version (optional) version of the API to be used
+     */
+    constructor (accessToken: string, serviceApi: ServiceApi, version = 'v1') {
```
kbridbur:
```suggestion
    constructor (accessToken: string, serviceApi: ServiceApi, version: string) {
```
We're the only ones who will be invoking this so I wonder if we should just make it explicit

src/base-api-client.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Get a list of jobs submitted within the last 30 days in reverse chronological order 
+     * (last submitted first) up to the provided limit number of jobs per call. Pagination is supported via passing
+     * the last job id from previous call into starting_after.
+     * @param params
+     * @returns List of job details
+     */
+    protected async _getListOfJobs(params?: {}): Promise<TJob[]> {
```
kbridbur, line 34: These params are consistent across apis and should always be. I think we can have them hardcoded here

kbridbur: I guess you did this for consistency with how the result call works. I prefer it the old way so we don't duplicate the logic of building this params dictionary in every implementation of this client

src/base-api-client.ts
```diff
@@ -0,0 +1,90 @@
+import { ApiRequestHandler } from './api-request-handler';
+
+type ServiceApi = 'topic_extraction' | 'sentiment_analysis';
```
kbridbur: I wonder if this is useful, I don't really see the point. If we're going to have it we should have one for version as well.

src/models/topic-extraction/TopicExtractionGetListOfJobsOptions.ts
```diff
@@ -0,0 +1,5 @@
+/** Topic extraction get list of jobs options */
+export interface TopicExtractionGetListOfJobsOptions {
```
kbridbur: This seems unnecessary and like we'll have unnecessary duplication or every api. Can we at least just call it GetListOfJobsOptions and put it in a place where every api using the new client can use it?

src/base-api-client.ts
kbridbur: Now that I left [this comment ](https://github.com/revdotcom/revai-node-sdk/pull/100/files#r878532135) I think it should look like this
```suggestion
    protected async _getListOfJobs(params?: GetListOfJobsOptions): Promise<TJob[]> {
```

src/models/topic-extraction/TopicExtractionJob.ts
```diff
@@ -0,0 +1,18 @@
...
+    type: JobType;
+    created_on: string;
+    completed_on?: string;
+    metadata?: string;
+    callback_url?: string;
+    delete_after_seconds?: number;
+    failure?: string;
+    failure_detail?: string;
+    word_count?: number;
+    language?: string;
```
kbridbur, line 20: Topic extraction job models do not include language
```suggestion
```

src/models/topic-extraction/TopicExtractionJobOptions.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Options that can used when submitting Rev AI topic extraction job.
+ * See https://docs.rev.ai/api/topic-extraction/reference/#operation/SubmitTopicExtractionJob for more details.
+ */
+export interface TopicExtractionJobOptions {
+    metadata?: string;
+    callback_url?: string;
+    delete_after_seconds?: number;
+    language?: string;
```
kbridbur: Language is not an option for this api

kbridbur: This made me realize that we need to go back and remove it from the python sdk

src/topic-extraction-client.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Client which handles connection to the Rev AI topic extraction API.
+ */
+export class TopicExtractionClient extends BaseApiClient<TopicExtractionJob, TopicExtractionResult> {
+    /**
+     * @param accessToken Access token used to validate API requests
+     * @param version (optional) version of the API to be used
+     */
+    constructor (accessToken: string, version = 'v1') {
```
kbridbur: No need to expose version to users. The sdks should use the latest version of the api

src/topic-extraction-client.ts
```diff
@@ -0,0 +1,48 @@
...
+        super(accessToken, 'topic_extraction', version);
+    }
+
+    async getJobDetails(id: string): Promise<TopicExtractionJob> {
```
kbridbur, line 29: xml docs? I don't think they'll be inherited from the base

test/unit/base-api-client.spec.ts
kbridbur: We've discussed this and it seems reasonable to me. I'd like @dmtrrk or @k-weng to look

kbridbur (changes requested):
Delete/Result tests need to be written and I'm not happy with the GetListOfJobs interface. Lots of minor comments otherwise

test/unit/base-api-client.spec.ts
dmtrrk: briefly:
- base client should be abstract
- we don't need tests for abstract class, but we should test only public contracts

k-weng: Aren't all of these methods effectively tested by `topic-extraction-client.spec.ts`?

src/base-api-client.ts
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Base client implementation. Intended to be extended by a specific client per API
+ */
+export class BaseApiClient<TJob, TResult> {
```
k-weng: If it's intended to be extended, should we add an abstract keyword?
```suggestion
 * Base client implementation. Intended to be extended by a specific client per API
 */
export abstract class BaseApiClient<TJob, TResult> {
```

src/base-api-client.ts
```diff
@@ -0,0 +1,90 @@
...
+    protected async _getListOfJobs(params?: {}): Promise<TJob[]> {
+        const query = this.buildQueryParams(params || {});
+        return await this.apiHandler.makeApiRequest<TJob[]>('get',
+            `/jobs${query ? `?${query}` : ''}`, {}, 'json');
+    }
+
+    /**
+     * Delete a specific job.
+     * All data related to the job will be permanently deleted.
+     * A job can only by deleted once it's completed.
```
beaudrychase, line 43: can failed jobs be deleted?
are failed jobs considered completed too?

hbkse: failed is a separate enum value but they can be deleted

examples/topic_extraction_example.js
hbkse: I changed it slightly to have pretty-print `console.log(JSON.stringify(topics, null, 2));`
In my console it looks like this
image
I think the result can be very large so it's not good to print the whole thing. I think it would be ok to leave this line as a comment

src/models/topic-extraction/TopicExtractionJob.ts
hbkse: I believe we do return language in the job details 
https://docs.rev.ai/api/topic-extraction/reference/#operation/SubmitTopicExtractionJob

The property is valid on submit also, but not used for anything currently, so I've removed it from the submit options object here. Unless, you think we should keep it as a reflection of our API docs

examples/topic_extraction_example.js
```diff
@@ -0,0 +1,51 @@
...
+    console.log('Polling for job completion...');
+
+    while((jobStatus = (await client.getJobDetails(job.id)).status) == revai.JobStatus.InProgress)
+    {
+        console.log(`Job ${job.id} is ${jobStatus}`);
+        await new Promise( resolve => setTimeout(resolve, 1000));
+    }
+
+    const topics = await client.getResult(job.id, { threshold: 0 });
+    // console.log(JSON.stringify(topics, null, 2));
```
kbridbur: No need to comment this line. Users should see their results by default

src/models/topic-extraction/TopicExtractionResult.ts
```diff
@@ -0,0 +1,20 @@
+/** Topic extraction result model*/
+export interface TopicExtractionResult {
+    topics: Topic[];
+}
+
+/** Topic extraction topic */
+export interface Topic {
+    topic_name: string;
```
kbridbur, line 11: xml docs on these properties. Would they be useful?

src/base-api-client.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Get a list of jobs submitted within the last 30 days in reverse chronological order
+     * (last submitted first) up to the provided limit number of jobs per call. Pagination is supported via passing
+     * the last job id from previous call into starting_after.
+     * @param params Query params for this request
+     * @returns List of job details
+     */
+    protected async _getListOfJobs(params?: {}): Promise<TJob[]> {
```
kbridbur, line 34:
```suggestion
    protected async _getListOfJobs(params?: GetListOfJobsOptions): Promise<TJob[]> {
```
Any reason not to have this?

kbridbur (approved):
A few comments but generally looks good

src/models/topic-extraction/TopicExtractionResult.ts
hbkse: yeah, I think so, I'll add

src/base-api-client.ts
hbkse: I see it where `GetListOfJobsOptions` is a user facing model and this class should be able to accept any object

examples/topic_extraction_example.js
```diff
@@ -0,0 +1,51 @@
...
+    // const job = await client.submitJobFromTranscript(transcript, jobOptions);
+
+    console.log('Topic extraction job submitted.');
+    console.log(`Job Id: ${job.id}`);
+    console.log('Polling for job completion...');
+
+    while((jobStatus = (await client.getJobDetails(job.id)).status) == revai.JobStatus.InProgress)
+    {
+        console.log(`Job ${job.id} is ${jobStatus}`);
+        await new Promise( resolve => setTimeout(resolve, 1000));
```
dmtrrk:
```suggestion
        await new Promise(resolve => setTimeout(resolve, 1000));
```

examples/topic_extraction_example.js
```diff
@@ -0,0 +1,51 @@
...
+    // const asyncApiClient = new revai.RevAiApiClient(token);
+    // const transcript = await asyncApiClient.getTranscriptObject(asyncJobId);
+    // console.log(`Pulling transcript from async job ${asyncJobId}...`);
+    // const job = await client.submitJobFromTranscript(transcript, jobOptions);
+
+    console.log('Topic extraction job submitted.');
+    console.log(`Job Id: ${job.id}`);
+    console.log('Polling for job completion...');
+
+    while((jobStatus = (await client.getJobDetails(job.id)).status) == revai.JobStatus.InProgress)
```
dmtrrk:
```suggestion
    while((jobStatus = (await client.getJobDetails(job.id)).status) === revai.JobStatus.InProgress)
```

dmtrrk (approved):
looks great. a few minor changes and good to go

PR #102, opened 2022-05-17
AISDK-205: Add Language Parameter to the Streaming Client

amikofalvy:


src/streaming-client.ts
```diff
@@ -104,6 +104,9 @@ export class RevAiStreamingClient extends EventEmitter {
             if (config.transcriber !== null && config.transcriber !== undefined) {
                 url += `&transcriber=${encodeURIComponent(config.transcriber)}`;
             }
+            if (config.language !== null && config.language !== undefined) {
```
dmtrrk:
```suggestion
            if (config.language) {
```

amikofalvy: Do you think we should update all of the conditionals?

dmtrrk (approved):
minor proposal

src/streaming-client.ts
dmtrrk: this might be a techdebt, so I approved this PR.
if it is non-trivial change, let's have a separate tech debt story

amikofalvy: It doesn't work for the numerical parameters such as delete_after_seconds because `0` is a `falsey` value.

dmtrrk: it works for language and everything what is not `0` - for any string values actually (except `"0"` string).

amikofalvy: Yes. I have updated the conditionals to reflect this. Please check again.

dmtrrk: looks great

PR #103, opened 2022-05-19
Stop running integration tests on branches

hbkse:


hbkse:
image

integration tests are skipped

kbridbur (approved):
We don't make release branches on these repositories so this is reasonable. Please wait for Andrew to also approve

amikofalvy (approved):
Please confirm that the integration tests block deployment of new packages prior to merge.

PR #105, opened 2022-05-25
AISDK-195: Language id should be supported in node sdk

seanlam8:


src/base-api-client.ts
```diff
@@ -1,5 +1,10 @@
+import * as FormData from 'form-data';
+import * as fs from 'fs';
+import { Readable } from 'stream';
+
 import { ApiRequestHandler } from './api-request-handler';
 
+const TWO_GIGABYTES = 2e9; // Number of Bytes in 2 Gigabytes
 /**
```
dmtrrk:
```suggestion

/**
```

src/base-api-client.ts
```diff
@@ -59,6 +64,47 @@ export abstract class BaseApiClient<TJob, TResult> {
...
+        options?: {}
+    ): Promise<TJob> {
+        const payload = new FormData();
+        payload.append('media', audioData, { filename: filename || 'audio_file' });
+        if (options) {
+            options = this.filterNullOptions(options);
+            payload.append('options', JSON.stringify(options));
+        }
+
+        return await this.apiHandler.makeApiRequest<TJob>('post', `/jobs`,
```
dmtrrk:
```suggestion
        return await this.apiHandler.makeApiRequest<TJob>('post', '/jobs',
```

src/base-api-client.ts
```diff
@@ -59,6 +64,47 @@ export abstract class BaseApiClient<TJob, TResult> {
...
+     */
+     async _submitJobLocalFile(filepath: string, options?: {}): Promise<TJob> {
+        const payload = new FormData();
+        payload.append('media', fs.createReadStream(filepath));
+        if (options) {
+            options = this.filterNullOptions(options);
+            payload.append('options', JSON.stringify(options));
+        }
+
+        return await this.apiHandler.makeApiRequest<TJob>('post', `/jobs`,
```
dmtrrk:
```suggestion
        return await this.apiHandler.makeApiRequest<TJob>('post', '/jobs',
```

src/language-id-client.ts
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Client which handles connection to the Rev AI language id API.
+ */
+export class LanguageIdClient extends BaseApiClient<LanguageIdJob, LanguageIdResult> {
+    /**
+     * @param accessToken Access token used to validate API requests
+     * @param version (optional) version of the API to be used
+     */
+     constructor (accessToken: string) {
```
dmtrrk: I see we are adding space after ctor word in other files. not sure was a mistake from the beginning. should that space be there?

seanlam8: oh yeah I'm not sure there should be, I copied the format from other clients, I can remove space

src/base-api-client.ts
```diff
@@ -59,6 +65,47 @@ export abstract class BaseApiClient<TJob, TResult> {
...
     }
 
+    /**
+     * Submit audio data for job submission.
+     * @param audioData Audio data to be submitted for job.
+     * @param filename (optional) Name of file associated with audio.
+     * @param options (optional) Options submitted with the job.
+     * @returns Details of submitted job
+     */
+     async _submitJobAudioData(
```
kbridbur:
```suggestion
     protected async _submitJobAudioData(
```
Same for the other method.

I'm generally a little concerned about these being on the base-api-client because it means the methods are available to the Topic and Sentiment clients as well which allows users to misuse them

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
...
+const token = require('./config/config.json').access_token;
+
+(async () => {
+    // Initialize your client with your Rev AI access token
+    var client = new revai.LanguageIdClient(token);
+
+    const jobOptions = {
+        metadata: 'node example language id local file submission',
+        notification_config: { url: 'https://jsonplaceholder.typicode.com/posts' },
+        delete_after_seconds: 2592000 // 30 days in seconds
```
hbkse:
```suggestion
        delete_after_seconds: 30 * 24 * 60 * 60 // 30 days in seconds
```

hbkse:
```suggestion
        delete_after_seconds: 30 * 24 * 60 * 60 // 30 days in seconds
```

src/base-api-client.ts
seanlam8: I can move them to language-id-client

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
+const revai = require('revai-node-sdk');
```
hbkse, line 1: I wonder if it's necessary in general to have 2 separate examples for local media file vs url submission. We could combine the two with sections commented out and leave it up to the users to comment in the parts they want. But, maybe this would make the file too large and confusing

gweng88, line 1: I wonder if it's necessary in general to have 2 separate examples for local media file vs url submission. We could combine the two with sections commented out and leave it up to the users to comment in the parts they want. But, maybe this would make the file too large and confusing

src/base-api-client.ts
hbkse: are they still access-able if they are marked as protected? we can do `topicClient._submitJobAudioData()`?

gweng88: are they still access-able if they are marked as protected? we can do `topicClient._submitJobAudioData()`?

examples/language_id_local_media_file.js
seanlam8: I guess I was just following the async example convention of having separate examples for local media and url. I think it is more clear this way

src/base-api-client.ts
seanlam8: I don't think they would be since only classes that inherit base-api-client could access them. But I moved the methods to language-id-client anyways

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,51 @@
...
+    console.log(`Status: ${job.status}`);
+    console.log(`Created On: ${job.created_on}`);
+
+    /**
+     * Waits 5 seconds between each status check to see if job is complete.
+     * note: polling for job status is not recommended in a non-testing environment.
+     * Use the notification_config option (see: https://docs.rev.ai/sdk/node/)
+     * to receive the response asynchronously on job completion
+     */
+    while((jobStatus = (await client.getJobDetails(job.id)).status) == revai.JobStatus.InProgress)
```
hbkse:
```suggestion
    while((jobStatus = (await client.getJobDetails(job.id)).status) === revai.JobStatus.InProgress) {
```

src/base-api-client.ts
```diff
@@ -6,7 +6,7 @@ import { ApiRequestHandler } from './api-request-handler';
 export abstract class BaseApiClient<TJob, TResult> {
     apiHandler: ApiRequestHandler;
 
-    /**
+    /**s
```
hbkse:
```suggestion
    /**
```

src/models/language-id/LanguageIdJobOptions.ts
```diff
@@ -0,0 +1,14 @@
+import { CustomerUrlData } from '../CustomerUrlData';
+
+/**
+ * Options that can used when submitting Rev AI language id job.
+ * See https://docs.rev.ai/api/language-identification/reference/#operation/SubmitLanguageIdentificationJob for more details.
+ */
+export interface LanguageIdJobOptions {
+    media_url?: string;
```
hbkse, line 8: If this is deprecated, I don't think we should be adding it to new APIs. @aaron-wilson-rev what are your thoughts?

src/language-id-client.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /**
+     * See https://docs.rev.ai/api/language-identification/reference/#operation/SubmitLanguageIdentificationJob
+     * Submits a language id job with a media url as input.
+     * @param mediaUrl Media url to be submitted for language id.
+     * @param options Options submitted with the job: see LanguageIdJobOptions object
+     * @returns Details of the submitted job
+     */
+    async submitJobUrl(mediaUrl: string, options: LanguageIdJobOptions = {}): Promise<LanguageIdJob> {
```
hbkse: @aaron-wilson-rev same here, if media_url is deprecated should we be allowing this?

hbkse (approved):
PR looks good to me, with the exception of adding `media_url`. Once that gets resolved it's good

src/models/language-id/LanguageIdJobOptions.ts
seanlam8: Actually I think as long as we keep it in the documentation, we need it here. Since it could still get used

src/language-id-client.ts
seanlam8: I changed it to accept source_config

src/models/language-id/LanguageIdJobOptions.ts
aaron-wilson-rev: Yeah I think we should have it here because the other sdks and our api still support it for language id. But if we add a new feature then we shouldn't continue adding it.

src/language-id-client.ts
aaron-wilson-rev: I don't think we need the submitJobUrl method at all for this client, since no one will be using it already. So here I think we could force usage of the submitJob with the job options set. If they still want to use media url in the job options they could for now but it would remove the url only submission option.

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
...
+    // Initialize your client with your Rev AI access token
+    var client = new revai.LanguageIdClient(token);
+
+    const jobOptions = {
+        metadata: 'node example language id local file submission',
+        notification_config: { url: 'https://jsonplaceholder.typicode.com/posts' },
+        delete_after_seconds: 30 * 24 * 60 * 60 // 30 days in seconds
+    };
+
+    // Media may be submitted from a local file
```
jennywong2129: not sure this comment is necessary

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
...
+        job = await client.submitJobLocalFile('./resources/example.mp3',
+            jobOptions);
+    } catch (e) {
+        console.dir(e);
+    }
+
+    console.log('Language id job submitted.');
+    console.log(`Job Id: ${job.id}`);
+    console.log(`Status: ${job.status}`);
+    console.log(`Created On: ${job.created_on}`);
```
jennywong2129: I would move these logs into the try block

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Waits 5 seconds between each status check to see if job is complete.
+     * note: polling for job status is not recommended in a non-testing environment.
+     * Use the notification_config option (see: https://docs.rev.ai/sdk/node/)
```
jennywong2129, line 29: @kbridbur is there a reason we dont link directly to the section 
https://docs.rev.ai/sdk/node/code-samples/#send-email-notifications-using-a-webhook

is it because that url could change/?

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
...
+     * to receive the response asynchronously on job completion
+     */
+    while((jobStatus = (await client.getJobDetails(job.id)).status) === revai.JobStatus.InProgress) {
+        console.log(`Job ${job.id} is ${jobStatus}`);
```
jennywong2129, line 33: not sure we need to parameterize job status

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
...
+     * Get language id result as Object
+     */
+     var languageIdResult = await client.getResult(job.id);
+     console.log(JSON.stringify(languageIdResult, null, 2));
+
+    /**
+     * Delete a job
+     * Job deletion will remove all information about the job from the servers
+     */
+    // await client.deleteJob(job.id);
```
jennywong2129, line 53: can we log deleted job id with id

examples/language_id_local_media_file.js
seanlam8: shouldn't we only log this if the job successfully submitted?

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
...
+     * note: polling for job status is not recommended in a non-testing environment.
+     * Use the notification_config option (see: https://docs.rev.ai/sdk/node/)
+     * to receive the response asynchronously on job completion
+     */
+    while((jobStatus = (await client.getJobDetails(job.id)).status) === revai.JobStatus.InProgress) {
+        console.log(`Job ${job.id} is ${jobStatus}`);
+        await new Promise( resolve => setTimeout(resolve, 5000));
+    }
+
+    /**
```
jennywong2129, line 37: we should log if the job failed, since we can't get results if failed

examples/language_id_local_media_file.js
```diff
@@ -0,0 +1,50 @@
+const revai = require('revai-node-sdk');
+const token = require('./config/config.json').access_token;
+
+(async () => {
+    // Initialize your client with your Rev AI access token
+    var client = new revai.LanguageIdClient(token);
+
+    const jobOptions = {
+        metadata: 'node example language id local file submission',
+        notification_config: { url: 'https://jsonplaceholder.typicode.com/posts' },
```
jennywong2129: not sure why we have this when in the bottom comment we ask user to read on this options, I think we should leave it out in the request her

examples/language_id_media_from_url.js
```diff
@@ -0,0 +1,59 @@
...
+    var client = new revai.LanguageIdClient(token);
+
+    // Configure your source media url
+    // If authorization headers are needed to access the url they can be provided as an argument, e.g.
+    // var sourceConfig  {url: 'source url', auth_headers: {"Authorization": "Bearer <token>"}});
+    const sourceConfig = { url: 'https://www.rev.ai/FTC_Sample_1.mp3', auth_headers: null };
+    // Set an optional notification url 
+    // See https://docs.rev.ai/api/asynchronous/webhooks/ for details on setting up a webhook
+    // Authorization headers url can also be added to this url, e.g. 
+    // var notificationConfig {url: 'webhook url', auth_headers: {"Authorization": "Bearer <token>"}});
```
jennywong2129, line 15: I think we should add the same comment for the user to read up on notification)config here as u did in the local suit examaple,
and maybe leave it off the sample request

examples/language_id_media_from_url.js
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    const jobOptions = {
+        source_config: sourceConfig,
+        metadata: 'node example language id url submission',
+        notification_config: notificationConfig,
+        delete_after_seconds: 30 * 24 * 60 * 60 // 30 days in seconds
+    };
+
+    // Media may be submitted from a source config
```
jennywong2129: comment seems unneeded

examples/language_id_media_from_url.js
```diff
@@ -0,0 +1,59 @@
...
+    var job = await client.submitJob(jobOptions);
+
+    console.log('Language id job submitted.');
+    console.log(`Job Id: ${job.id}`);
+    console.log(`Status: ${job.status}`);
+    console.log(`Created On: ${job.created_on}`);
```
jennywong2129: I think we ned 29-34 in a try catch like the local sample

examples/language_id_media_from_url.js
```diff
@@ -0,0 +1,59 @@
...
+     */
+    var languageIdResult = await client.getResult(job.id);
+    console.log(JSON.stringify(languageIdResult, null, 2));
+
+    /**
+     * Delete a job
+     * Job deletion will remove all information about the job from the servers
+     */
+    // await client.deleteJob(job.id);
+})();
```
jennywong2129, line 69: same comments as I had on the other sample file about not needing to parameterize status in 43, logging if status is failed/sucess and comment for deleted Job id

src/language-id-client.ts
```diff
@@ -0,0 +1,123 @@
...
+import { LanguageIdJob } from './models/language-id/LanguageIdJob';
+import { LanguageIdJobOptions } from './models/language-id/LanguageIdJobOptions';
+import { LanguageIdResult } from './models/language-id/LanguageIdResult';
+
+const LanguageIdContentType = 'application/vnd.rev.languageid.v1.0+json';
+
+const TWO_GIGABYTES = 2e9; // Number of Bytes in 2 Gigabytes
+
+/**
+ * Client which handles connection to the Rev AI language id API.
```
jennywong2129: capitalized language id

src/language-id-client.ts
```diff
@@ -0,0 +1,123 @@
...
+     * @param options Options submitted with the job: see LanguageIdJobOptions object
+     * @returns Details of the submitted job
+     */
+     async submitJob(options: LanguageIdJobOptions = {}): Promise<LanguageIdJob> {
+        return super._submitJob(options);
+    }
+
+    /**
+     * See https://docs.rev.ai/api/language-identification/reference/#operation/SubmitLanguageIdentificationJob
+     * Submit audio data for language id.
```
jennywong2129: can we say from local file, and the method above from media  url

src/language-id-client.ts
```diff
@@ -0,0 +1,123 @@
...
+            payload.append('options', JSON.stringify(options));
+        }
+
+        return await this.apiHandler.makeApiRequest<LanguageIdJob>('post', `/jobs`,
+            payload.getHeaders(), 'json', payload, TWO_GIGABYTES);
+    }
+
+    /**
+     * See https://docs.rev.ai/api/language-identification/reference/#operation/GetLanguageIdentificationResultById
+     * Get the result of a language id job.
```
jennywong2129: of a successful language id job

src/models/language-id/LanguageIdJob.ts
```diff
@@ -0,0 +1,18 @@
...
+export interface LanguageIdJob {
+    id: string;
+    status: JobStatus;
+    type: JobType;
+    created_on: string;
+    completed_on?: string;
+    metadata?: string;
+    callback_url?: string;
+    processed_duration_seconds?: number;
+    media_url?: string;
```
jennywong2129, line 14: we shouldn't return media url/callback url right?

seanlam8: I think while it is still supported in api and docs, we should keep them here

test/integration/test/language-id/get-job.test.js
```diff
@@ -0,0 +1,35 @@
...
+        options2.metadata = 'Node sdk submit url 2';
+        await client.submitJobUrl('https://www.rev.ai/FTC_Sample_1.mp3', options1);
+        await client.submitJobUrl('https://www.rev.ai/FTC_Sample_1.mp3', options2);
+    }
+    done();
+}, 60000);
+
+test('Can get list of jobs', async () => {
+    const jobList = await client.getListOfJobs();
+    jobList.forEach((languageIdJob) => {
```
jennywong2129, line 24: can we loop with an index to check that we see metadata for url 2 before url 1

test/integration/test/language-id/get-job.test.js
```diff
@@ -0,0 +1,35 @@
...
+        expect(languageIdJob).toMatchObject(LanguageIdJob);
+    });
+}, 30000);
+
+test('Can get single job and job by id', async () => {
+    const jobList = await client.getListOfJobs(1);
+    expect(jobList.length).toEqual(1);
+    expect(jobList[0]).toMatchObject(LanguageIdJob);
```
jennywong2129, line 35: check meta is url 2

test/integration/test/language-id/submit.test.js
```diff
@@ -0,0 +1,68 @@
...
+const JobStatus = require('../../../../dist/src/models/JobStatus').JobStatus;
+const client = clientHelper.getLanguageIdClient();
+
+test('Can submit job', async () => {
+    const options = new Object();
+    options.metadata = 'Node sdk submit url';
+    options.sourceConfig = { url: 'https://www.rev.ai/FTC_Sample_1.mp3', auth_headers: null };
+    const res = await client.submitJob(options);
+
+    expect(res.status).toBe(JobStatus.InProgress);
```
jennywong2129, line 13: res.type = language_id

jennywong2129 (approved):
looks good except comments can be refined, Great Job! :shipit:

examples/language_id_local_media_file.js
jennywong2129: yea, so try will have the submit + all logs, catch will log the error, you had it already in the url sample file

examples/language_id_local_media_file.js
seanlam8: I think it looks cleaner this way also couldn't the value change if we call `getJobDetails` twice

examples/language_id_media_from_url.js
seanlam8: Changing this to `Submitting job via source config`

examples/language_id_media_from_url.js
seanlam8: We link to webhooks, which explains notification config

seanlam8: I think we should have an example of submitting with notification config

test/integration/test/language-id/get-job.test.js
seanlam8: We can't since we submit other jobs for the other language id integration tests, so when we `getListOfJobs` we get all of those ones too.

test/integration/test/language-id/get-job.test.js
seanlam8: After running integration tests locally, I realized that these jobs get submitted in parallel it seems, so we can't guarantee this.

examples/language_id_local_media_file.js
jennywong2129: fair enough

test/integration/test/language-id/get-job.test.js
jennywong2129: then we should check decreasing createdOn times

test/integration/test/language-id/get-job.test.js
jennywong2129: that's ok then

PR #108, opened 2022-06-02
AISDK-221: Fix HIPAA Error Message on Incorrect Usage

aaron-wilson-rev:
Submission to the API with insufficient balance does not actually trigger a 403. The job submits successfully then fails. I also changed the error field as it looks like that wasn't ever populating. Now using the wrong parameter as HIPAA will return the right error.

### DIFF
--- a/src/api-request-handler.ts
+++ b/src/api-request-handler.ts
@@ -63,14 +63,10 @@ export class ApiRequestHandler {
             switch (error.response.status) {
                 case 400:
                     throw new InvalidParameterError(error);
-                case 401:
-                case 403:
-                case 404:
-                    throw new RevAiApiError(error);
                 case 409:
                     throw new InvalidStateError(error);
                 default:
-                    throw error;
+                    throw new RevAiApiError(error);
             }
         }
     }
--- a/src/api-request-handler.ts
+++ b/src/api-request-handler.ts
@@ -68,7 +68,11 @@ export class ApiRequestHandler {
                 case 404:
                     throw new RevAiApiError(error);
                 case 403:
-                    throw new InsufficientCreditsError(error);
+                    if (error.response.data.error.includes("Not enough credits for payment")) {
+                        throw new InsufficientCreditsError(error);
+                    } else {
+                        throw new RevAiApiError(error)
+                    }
                 case 409:
                     throw new InvalidStateError(error);
                 default:
--- a/src/api-request-handler.ts
+++ b/src/api-request-handler.ts
@@ -65,14 +65,9 @@ export class ApiRequestHandler {
                 case 400:
                     throw new InvalidParameterError(error);
                 case 401:
+                case 403:
                 case 404:
                     throw new RevAiApiError(error);
-                case 403:
-                    if (error.response.data.error.includes("Not enough credits for payment")) {
-                        throw new InsufficientCreditsError(error);
-                    } else {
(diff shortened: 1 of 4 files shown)


src/api-request-handler.ts
```diff
@@ -68,7 +68,11 @@ export class ApiRequestHandler {
                 case 404:
                     throw new RevAiApiError(error);
                 case 403:
-                    throw new InsufficientCreditsError(error);
+                    if (error.response.data.error.includes("Not enough credits for payment")) {
```
aaron-wilson-rev: I'm not sure this ever actually gets triggered. It seems like now the submission still succeeds, the job will just fail. I don't know if it's always been this way...

src/models/RevAiApiError.ts
```diff
@@ -3,14 +3,14 @@ import { AxiosError } from 'axios';
...
+    error?: string;
     type?: string;
 
     constructor(e: AxiosError) {
         if (e.response) {
             this.statusCode = e.response.status;
             this.title = e.response.data.title || '';
-            this.detail = e.response.data.detail || '';
+            this.error = e.response.data.error || '';
             this.type = e.response.data.type || '';
```
aaron-wilson-rev: I'm also not sure why we're trying to filter out what's important here. Why not just show the user just response.status and response.data?

src/api-request-handler.ts
```diff
@@ -65,10 +64,9 @@ export class ApiRequestHandler {
                 case 400:
                     throw new InvalidParameterError(error);
                 case 401:
+                case 403:
                 case 404:
                     throw new RevAiApiError(error);
-                case 403:
```
dmtrrk, line 70: I have a feeling that we're using this scenario for multiple reasons.
Check [Python SDK](https://github.com/revdotcom/revai-python-sdk/blob/447ed2a1aa73289df5c1a44fbfea09319cb0b5ab/tests/helpers/errors.py#L30)

@kbridbur do you have any ideas on this?

aaron-wilson-rev: Submitting a request as a user with no credits does not actually return a 403 so wrapping it in a specific credits error is never used; that's why I removed it here.

aaron-wilson-rev: I guess that python sdk error should be removed as well -- it's never used

src/models/RevAiApiError.ts
aaron-wilson-rev: It makes it difficult to keep updated if the specific field changes. I'm guessing that's what happened here and the sdks never got updated and we switched to using "error" instead of "detail".

src/api-request-handler.ts
dmtrrk: I don't see anything in the code that returns 403. or `out-of-credit`. I think that was a previous implementation.
Let's keep this as is. Please post Python SDK ticket to make it consistent with your update

src/models/RevAiApiError.ts
aaron-wilson-rev: Went ahead and applied this change

eugenep-rev (approved):
Just curious, how do you test it?

gvaughnRev (approved):
This looks reasonable to me

aaron-wilson-rev:
> Just curious, how do you test it?

Node SDK can be tested easily by just running or modifying the examples included with my access token in the example folder

PR #124, opened 2022-06-09
AISDK-189:  Update all clients to have the option to submit to eu endpoint

jennywong2129:


examples/async_transcribe_local_media_file.js
```diff
@@ -4,7 +4,8 @@ const token = require('./config/config.json').access_token;
 
 (async () => {
     // Initialize your client with your Rev AI access token
-    var client = new revai.RevAiApiClient(token);
+    // Make sure the base url matches the deployment for your Rev AI access token
+    var client = new revai.RevAiApiClient(accessToken: token, version: 'v1', baseUrl: BaseUrl.US);
```
dmtrrk: did we hide the base url intentionally before this change?

examples/async_transcribe_media_from_url.js
```diff
@@ -4,7 +4,8 @@ const token = require('./config/config.json').access_token;
 
 (async () => {
     // Initialize your client with your Rev AI access token
-    var client = new revai.RevAiApiClient(token);
+    // Make sure the base url matches the deployment for your Rev AI access token
+    var client = new revai.RevAiApiClient(accessToken: token, version: 'v1', baseUrl: BaseUrl.US);
```
dmtrrk: I propose not to expose `baseUrl` for the US examples, so it is US by default

src/api-client.ts
```diff
@@ -25,9 +26,11 @@ export class RevAiApiClient {
     /**
      * @param accessToken Access token used to validate API requests
      * @param version (optional) version of the API to be used
+     * @param baseUrl (optional) base url of the API to be used.
+     *     Note that non-US global deployments have a different base url
      */
-    constructor (accessToken: string, version = 'v1') {
-        this.apiHandler = new ApiRequestHandler(`https://api.rev.ai/speechtotext/${version}/`, accessToken);
+    constructor (accessToken: string, version = 'v1', baseUrl = BaseUrls.US) {
```
dmtrrk: my concern with this approach is that you can't have (easily) named parameters, so to specify the `baseUrl` you have to specify the version.

```
new Client(accessToken = '', baseUrl = '');
```

I propose to introduce a new (additional) constructor e.g.
```
interface SomeMagicApiConfig {
  accessToken: string;
  version?: string;
  baseUrl?: string;
}

constructor(SomeMagicApiConfig  config) {
  const mergedConfig = Object.assign(this.defaultConfig, config);
}

const client = new Client({ accessToken: '', baseUrl: '' });
```

in this case users can specify whatever they need without the additional headache.

cc: @k-weng  @amikofalvy

examples/async_transcribe_local_media_file.js
jennywong2129: im not sure intentionally, there was only 1 base url so there was no need to offer the option to set it I guess @amikofalvy

src/api-client.ts
jennywong2129: I think this would be a huge breaking change

dmtrrk: this goes in addition to the current ctor so there is no breaking change.
we will use the new ctor in the old one, but any new parameters will be added to the new ctor config only

examples/async_transcribe_local_media_file.js
amikofalvy: When we created this SDK originally there was no need to allow the user to change the base url so we did not expose it as part of the interface to reduce the complexity.

Interestingly, we did want this functionality for our integration tests so that we could run them against non production environments:

https://github.com/revdotcom/revai-node-sdk/blob/8bca62539300fdfb05d1ee7086207b7a0cd05597/test/integration/src/client-helper.js#L12-L16

I think you should update the client-helper to use the client constructor parameter instead of overriding the object property.

src/api-request-handler.ts
```diff
@@ -63,6 +65,10 @@ export class ApiRequestHandler {
             switch (error.response.status) {
                 case 400:
                     throw new InvalidParameterError(error);
+                case 403:
+                    throw new ForbiddenAccessError(error);
+                case 404:
```
amikofalvy, line 70: Where is the 403 status code documented on docs.rev.ai?

I don't see anything on https://docs.rev.ai/api/asynchronous/reference/#operation/GetJobById


Also, 404 is an expected status code for GetJobById

amikofalvy (changes requested):
Let's have a short talk on error codes on Thursday either on Zoom or over Slack.

examples/async_transcribe_local_media_file.js
dmtrrk: Yes, I was wondering what was the purpose when tried to test the client across multiple environments.
We agreed to introduce a new constructor that will stay fully compatible with the existing code, but is more extensible.

Example:
https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMIHsQ2Ac2QbwChkTkx0BrCEAfgC5kBnMKUHAbmNIDdpHhM9JizadSyALZwcwBEJABXCQCNonAL6FCCADZxGjNDuDUwBLiQAOrbnEjIA+gkzYcDDFlzIAvAXWcLZGcQZigFBHIoAApg13cXLwAfYVYQHAAaZF4ofkwGULYfZABybgBGYoBKQKJxcWAYZCiwAE9LCHRG2K9vXpL0ZQArCAiq8zqJsgALYEYAOicEvF8AeSGRsDn9fhwQZpn5xc8MoKXkfTQlyrFJ8TADhe6cOezckGQaGl9XgRAbifUyAgOkYKFqt1I91mjyWc3IVHevie-whUMOTxefF+RR+mBR4k0BK04msoDAUUq40mwUY6B0EDmOnQOH20KOrmugU0mkIAG1AiAIAB3IwmcBRfBkSjUBjFMoAJgAzMVkOpKukBcLRaYosV4dQFcr1ZqRahjDq9dKQIbiplSvKqoQALpzGDoKAAUUQUyiKG8AD4gXNSeLKpzCEA

src/api-client.ts
```diff
@@ -4,30 +4,55 @@ import { Readable } from 'stream';
...
 import { ApiRequestHandler } from './api-request-handler';
 import { CaptionType } from './models/async/CaptionType';
+import { TranscriptType } from './models/async/TranscriptType';
 import { RevAiAccount } from './models/async/RevAiAccount';
 import { RevAiJobOptions } from './models/async/RevAiJobOptions';
+import { RevAiBaseUrl } from './models/RevAiBaseUrl';
 import { RevAiApiJob } from './models/RevAiApiJob';
 import { RevAiApiTranscript } from './models/RevAiApiTranscript';
 
-const enum TranscriptContentTypes {
```
jennywong2129, line 13: moved into its own class

README.md
```diff
@@ -29,7 +29,13 @@ import { RevAiApiClient } from 'revai-node-sdk';
...
 // Initialize your client with your Rev AI access token
 const accessToken = "Your Access Token";
-const client = new RevAiApiClient(accessToken);
+/**
+ * Optionally set the base url of the Rev AI deployment of your account, defaults to US.
+ * Note the base url for the streaming API is different from the asynchronouse API,
+ * use RevAiBaseWebsocketUrl and RevAiBaseUrl, respectively.
+ * Learn more about Rev AI's global deployments [here](https://docs.rev.ai/api/global-deployments/).
+ */
+const client = new RevAiApiClient({ token: accessToken, baseUrl: RevAiBaseUrl.US } );
```
menioa: I don't think we should have `baseUrl` as the interface here. Customer's don't care about what url they are submitting to, but rather what region they are submitting to. Or if region is too specific, it could be something like deployment

src/models/RevAiApiClientConfig.ts
```diff
@@ -0,0 +1,16 @@
...
+ *    associated with the url in @param baseUrl
+ * @param version version of the API to be used
+ * @param baseUrl base url of the API to be used. Note base urls are different for Rev AI deployments in
+ *     different locations. Base urls are also different for asynchronous and streaming APIs.
+ * @param serviceApi Type of api service
+ */
+export interface RevAiApiClientConfig {
+    token?: string;
+    version?: string;
+    baseUrl?: string;
```
menioa: I think this should be region as I stated above. Additionally, this would allow us to handle url switching for async and streaming internally, rather than users needing to configure different url.

menioa (changes requested):
We should not be exposing BaseUrl to customers imo, I think instead we should have them specify deployment region

README.md
alexsku: should we just make  BaseUrls available to customers and recommend using those constant? I personally actually like flexibility of specifying your own url if needed

menioa: perhaps, but I think its a bit of an implementation leak. Take aws clients for example, you specify the region, you want, not the url. I think its more useful for streaming as well, so users don't have to swap out urls, but we can handle the switching internally

menioa: I'm happy to discuss if you want

jennywong2129: I think base url gives us more flexibility if in the future we have more than 1 deployment per region, or if a customer has a deployment just for them

menioa: I think in this case we could do `RevAiDeployment.US_1, RevAiDeployment.Customer`. I guess for the customer case, we probably wouldn't want to add a constant for either strategy, so they'd either have to modify internals or pass the prefix/url explicitly. In this case maybe we should do url explicitly, but I worry we may be putting the cart before the horse, as we still do not have any customer specific deployments.

alexsku (approved):
looks good to me, Albert will give final shipit

src/models/RevAiApiClientConfig.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param token Access token used to validate API requests, note access token should match the deployment
+ *    associated with the url in @param baseUrl
+ * @param version version of the API to be used
+ * @param deploymentConfig deployment configurations of the API to be used
+ * @param serviceApi Type of api service
+ */
+export interface RevAiApiClientConfig {
+    token?: string;
+    version?: string;
+    deploymentConfig?: RevAiAPiDeploymentConfig;
```
menioa:
```suggestion
    deploymentConfig?: RevAiApiDeploymentConfig;
```

src/models/RevAiApiClientConfig.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+/* eslint-disable no-shadow */
+/**
+ * Rev AI Deployemnt Configuration object to initialize RevAiApiClientConfig with.
+ * Note configurations are different for Rev AI deployments in different locations.
+ * @param baseUrl base url of the API to be used
+ * @param baseWebsocketUrl base urls of the streaming API to be used
+ */
+interface RevAiAPiDeploymentConfig {
```
menioa:
```suggestion
interface RevAiApiDeploymentConfig {
```

src/models/RevAiApiDeploymentConfigConstants.ts
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable no-shadow */
```
menioa, line 1: is this necessary?

menioa (approved):
some typos, but otherwise :shipit:

src/models/RevAiApiDeploymentConfigConstants.ts
jennywong2129: it fails eslint for this export otherwise

PR #129, opened 2022-06-22
AISDK-227: Replace TSLint with ESLint

amikofalvy:


.github/workflows/build_test.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x, 14.x, 16.x, 17.x]
```
amikofalvy, line 23: ESLint requires a minimum Node Version of 12.

All previous versions are beyond End Of Life and I don't think we need to continue to support them. In fact 12 EOL was over 1 month ago today. We can add additional modern versions in a separate ticket.

package.json
```diff
@@ -42,13 +43,17 @@
   "devDependencies": {
     "@types/form-data": "^2.2.1",
     "@types/jest": "^23.3.13",
-    "@types/node": "^10.12.18",
-    "@types/websocket": "^1.0.4",
+    "@types/node": "^12.20.55",
```
amikofalvy, line 46: I upgraded a few of the packages to work nicely with ESLint.

src/api-client.ts
```diff
@@ -1,6 +1,7 @@
-import * as FormData from 'form-data';
+/* eslint-disable no-shadow */
```
amikofalvy, line 1: In the migration guide, it recommended adding these to the files first then refining the eslint rules later as time goes on.

src/api-client.ts
```diff
@@ -97,7 +98,7 @@ export class RevAiApiClient {
             ...(options || {})
         });
 
-        return await this.apiHandler.makeApiRequest<RevAiApiJob>('post', `/jobs`,
+        return await this.apiHandler.makeApiRequest<RevAiApiJob>('post', '/jobs',
```
amikofalvy, line 101: I implemented a few of the auto-fixes.

src/api-request-handler.ts
```diff
@@ -9,7 +11,6 @@ import {
 export type HttpMethodTypes = 'post' | 'get' | 'delete';
 export type AxiosResponseTypes = 'stream' | 'json' | 'text';
 
-// tslint:disable-next-line
```
amikofalvy, line 12: Removed these annotations since tslint isn't being used anymore.

src/streaming-client.ts
```diff
@@ -55,10 +54,14 @@ export class RevAiStreamingClient extends EventEmitter {
             readableObjectMode: true,
             writableObjectMode: true
         });
-        this.client = new client({
-            // @ts-ignore
+        const clientConfig: IClientConfig = {};
+        const clientConfigExtensionProperties = {
```
amikofalvy, line 58: I liked the interface separation. The compiler was complaining after I updated the typescript version.

test/integration/test/transcript.test.js
```diff
@@ -1,28 +1,27 @@
 const clientHelper = require('../src/client-helper');
-const configHelper = require('../src/config-helper');
 const JobStatus = require('../../../dist/src/models/JobStatus').JobStatus;
 const client = clientHelper.getAsyncClient();
 
 beforeAll(async (done) => {
     const jobList = await client.getListOfJobs();
-    var jobId;
+    let jobId;
```
amikofalvy, line 7: more auto-fixes. we can refine the eslint rules later.

test/unit/api-client/api-client.submit.spec.ts
```diff
@@ -254,9 +254,9 @@ describe('api-client job submission', () => {
                 '_boundary': expect.anything(),
                 '_streams': expect.arrayContaining([
                     expect.anything(),
-                    expect.stringContaining('Content-Disposition: form-data; name="media"; filename=\"audio_file\"'),
```
amikofalvy, line 257: escape chars were not used according to eslint.

src/custom-vocabularies-client.ts
```diff
@@ -44,7 +44,7 @@ export class RevAiCustomVocabulariesClient {
             throw Error('customVocabularies is a required parameter');
         }
 
-        let options: CustomVocabularyOptions = {custom_vocabularies: customVocabularies};
+        const options: CustomVocabularyOptions = {custom_vocabularies: customVocabularies};
```
dmtrrk:
```suggestion
        const options: CustomVocabularyOptions = { custom_vocabularies: customVocabularies };
```

test/unit/sentiment-analysis-client.spec.ts
```diff
@@ -1,4 +1,9 @@
-import { SentimentAnalysisClient, SentimentAnalysisJob, SentimentAnalysisJobOptions, SentimentAnalysisResult, SentimentAnalysisResultOptions } from '../../src';
+import {
+    SentimentAnalysisClient,
+    SentimentAnalysisJob,
+    SentimentAnalysisJobOptions,
+    SentimentAnalysisResult,
+    SentimentAnalysisResultOptions } from '../../src';
```
dmtrrk:
```suggestion
    SentimentAnalysisResultOptions
} from '../../src';
```

test/unit/streaming-client.spec.ts
```diff
@@ -198,15 +197,15 @@ describe('streaming-client', () => {
...
+            const mockConnection = new WebSocketConnectionMock();
             sut.on('connect', response => jobId = response.id);
 
             // Act
             mockClient.emit('connect', mockConnection);
             mockConnection.emit('message',
                 {
                     type: 'utf8',
-                    utf8Data: `{ \"type\": \"connected\", \"id\": \"${expectedJobId}\"}`
+                    utf8Data: `{ "type": "connected", "id": "${expectedJobId}"}`
```
dmtrrk:
```suggestion
                    utf8Data: `{ "type": "connected", "id": "${expectedJobId}" }`
```

dmtrrk (approved):
a few minor changes

amikofalvy:
@dmtrrk @k-weng There is a question about how to version our SDK when we drop support for certain Node.js versions.

Originally, I was considering that this is a breaking change for our users therefore it requires a major version update even if it is backward compatible for most users who are already using LTS Versions of Node.

However, @k-weng pointed out that there are many examples of organizations which do not bump major versions when they drop Node support.

AWS: https://github.com/aws/aws-sdk-js-v3/blob/main/CHANGELOG.md#features-72
GatsbyJS: https://www.gatsbyjs.com/docs/upgrading-node-js/#gatsbys-nodejs-support-policy

These two example end support for EOL Node versions in minor versions which does not explicitly follow semantic versioning.

.github/workflows/build_test.yml
dmtrrk: This is a dependency for testing purpose. at the same time, I used `ls-engines` to check the dependencies graph and it shows `>= 6.14`. It takes all dependencies (not devDeps) and checks the graph
Why would we change the compatibility right now (we can add `ls-engines` verification to the pipeline, etc) as the package did not change as a fact?

amikofalvy: I followed your suggestion and only removed the out of date node versions from the lint step.

PR #227, opened 2023-12-22
REVAI-3853: usage-examples

kirillatrev:
REVAI-3853: usage-examples

### DIFF
--- a/examples/async_summarize_local_media_file.js
+++ b/examples/async_summarize_local_media_file.js
@@ -1,0 +1,72 @@
+const revai = require('revai-node-sdk');
+const fs = require('fs');
+const { JobStatus } = require('../dist/src/models/JobStatus');
+const { SummarizationJobStatus } = require('../dist/src/models/async/SummarizationJobStatus');
+const { SummarizationFormattingOptions } = require('../dist/src/models/async/SummarizationFormattingOptions');
+const token = require('./config/config.json').access_token;
+
+(async () => {
+    // Initialize your client with your Rev AI access token
+    var client = new revai.RevAiApiClient({ token: token });
+
+    // Get account details
+    var account = await client.getAccount();
+    console.log(`Account: ${account.email}`);
+    console.log(`Credits remaining: ${account.balance_seconds} seconds`);
+
+    const jobOptions = {
+        metadata: 'InternalOrderNumber=123456789',
+        delete_after_seconds: 2592000, // 30 days in seconds
+        language: 'en', // Supported ISO 639-1 (2-letter) or ISO 639-3 (3-letter) language code
+        summarization_config: {
+            type: SummarizationFormattingOptions.Paragraph
+        }
+    
+    };
+
+    // Media may be submitted from a local file
+    var job;
+    try {
+        job = await client.submitJobLocalFile('resources/example.mp3',
+            jobOptions);
+    } catch (e) {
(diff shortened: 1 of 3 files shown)


amikofalvy (approved):

examples/async_translate_local_media_file.js
```diff
@@ -0,0 +1,78 @@
...
+        if (err) throw err;
+        console.log('Success! Check the examples/outputs/ directory for the transcript.')
+    });
+
+    /**
+     * Delete a job
+     * Job deletion will remove all information about the job from the servers
+     */
+    // await client.deleteJob(job.id);
+})();
```
alexsku, line 78: why is this connected out? if we don’t need it i would remove it and if we need-uncomment

alexsku (approved):
:shipit:

examples/async_summarize_local_media_file.js
```diff
@@ -0,0 +1,71 @@
...
+const { SummarizationJobStatus } = require('../dist/src/models/async/SummarizationJobStatus');
+const { SummarizationFormattingOptions } = require('../dist/src/models/async/SummarizationFormattingOptions');
+const token = require('./config/config.json').access_token;
+
+(async () => {
+    // Initialize your client with your Rev AI access token
+    var client = new revai.RevAiApiClient({ token: token });
+
+    // Get account details
+    var account = await client.getAccount();
```
dmtrrk:
```suggestion
    const account = await client.getAccount();
```

examples/async_summarize_local_media_file.js
```diff
@@ -0,0 +1,71 @@
...
+    console.log(`Summarization status: ${job.summarization.status}`);
+    console.log(`Created On: ${job.created_on}`);
+
+    /**
+     * Waits 5 seconds between each status check to see if job is complete.
+     * note: polling for job status is not recommended in a non-testing environment.
+     * Use the notification_config option (see: https://docs.rev.ai/sdk/node/)
+     * to receive the response asynchronously on job completion
+     */
+    while((job = await client.getJobDetails(job.id))  && (
```
dmtrrk:
```suggestion
    while((job = await client.getJobDetails(job.id)) && (
```

examples/async_summarize_local_media_file.js
```diff
@@ -0,0 +1,71 @@
...
+        ))
+    {
+        console.log(`Job ${job.id} is ${job.status}, summarization is ${job.summarization.status}`);
+        await new Promise( resolve => setTimeout(resolve, 5000));
+    }
+
+    /**
+     * Get transcript summary as plain text
+     */
+    var transcriptSummary = await client.getTranscriptSummaryText(job.id);
```
dmtrrk:
```suggestion
    const transcriptSummary = await client.getTranscriptSummaryText(job.id);
```

examples/async_translate_local_media_file.js
kirillatrev: Some of the code in the existing example files is there just as an example and is commented out.

dmtrrk (approved):
Please check var vs const, indentation
Also, check for variable hoisting cases, eg `var job`. I think it is redundant

kirillatrev:
@dmtrrk hoisting - I think all examples follow the same pattern. Also I believe this is standard in JS, unless I totally missed your point....

PR #230, opened 2024-01-03
REVAI-3917: Split NlpModel to have separate enums for Translation and Summarization

kirillatrev:
REVAI-3917: Split NlpModel to have separate enums for Translation and Summarization
https://revinc.atlassian.net/browse/REVAI-3917

### DIFF
--- a/src/api-client.ts
+++ b/src/api-client.ts
@@ -318,12 +318,8 @@ export class RevAiApiClient {
      */
     async getTranslatedCaptions(id: string,
         language: string,
-        contentType?: string,
-        channelId?: number): Promise<Readable> {
+        contentType?: string): Promise<Readable> {
         let url = `/jobs/${id}/captions/translation/${language}`;
-        if (channelId) {
-            url += `?speaker_channel=${channelId}`;
-        }
         return await this.apiHandler.makeApiRequest<Readable>(
             'get', url, { 'Accept': contentType || CaptionType.SRT }, 'stream');
     }
--- a/src/api-client.ts
+++ b/src/api-client.ts
@@ -319,7 +319,7 @@ export class RevAiApiClient {
     async getTranslatedCaptions(id: string,
         language: string,
         contentType?: string): Promise<Readable> {
-        let url = `/jobs/${id}/captions/translation/${language}`;
+        const url = `/jobs/${id}/captions/translation/${language}`;
         return await this.apiHandler.makeApiRequest<Readable>(
             'get', url, { 'Accept': contentType || CaptionType.SRT }, 'stream');
     }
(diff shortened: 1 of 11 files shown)


src/api-client.ts
```diff
@@ -318,12 +318,8 @@ export class RevAiApiClient {
      */
     async getTranslatedCaptions(id: string,
         language: string,
-        contentType?: string,
-        channelId?: number): Promise<Readable> {
```
dmtrrk, line 322: would you pls add some comment to the ticket - why do we remove `channelId`?

src/models/async/TranslationMode.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable no-shadow */
+/** Supported model types for translation. */
+export enum TranslationMode {
```
dmtrrk: why do we emphasize that for Summarization we use a Model, but for the translation we have a Mode.

kirillatrev: Consistency with the backend inconsistencies :-)))
@alexsku okay to rename to TranslationModel ?

src/api-client.ts
kirillatrev: Just a piggyback cleanup since we need to update all 3 SDKs...

src/models/async/TranslationMode.ts
dmtrrk: btw it looks like the backend has `Fast` mode also

kirillatrev: We're not supporting Fast at this point.
I checked with @alexsku and he suggested not removing it from the backend for now.

dmtrrk: why is it a Model for Translation, but NLP Model for Summarization? why is it not a `SummarizationModel` instead of `NlpModel`?

kirillatrev: Historically this is how it was defined for Nexus.
Do you think it is worth renaming it at this point?

kirillatrev: Actually SummarizationModel is already in use...

dmtrrk: I think it does not matter how do we call it internally. from my understanding this is the first time we are exposing such a type. and it is always easier to have a tech debt to rename our internal code rather than making changes to the public API - users will hate us.

originally, it was `NlpModel` - some abstract model type offered by all NLP-related features (summarization, translation).
now, we want to make it more specific. so we have `TranslationModel` for the translation, which is pretty much specific. but `NlpModel` for Summarization. what if we have a topic extraction or any other API that can use different models? how would we call it?

kirillatrev: Don't get me wrong, I agree :-)
We can easily rename it in NodeJS SDK, it's just a bit more tricky in the backend.

dmtrrk (approved):

PR #291, opened 2024-12-02
feat: make websocket an optional peer dependency

attehuhtakangas:
This PR makes the `websocket` dependency a peer dependency. 

## The motivation for this change
- The `websocket` dependency is not used in all cases, only with the streaming client
- `websocket` dependency is arguably worse than `ws` and `websocket` pulls in a lot of native dependencies. It's a common practice that any dependency that pulls in `websocket` has it as an optional peer dependency for that given reason.

Additionally, this PR adds an `engines` definition and a `packageManager` field to make the expected tools easier to discover.


Hainesy (approved):

attehuhtakangas:
Not stale

